PR #2, opened 2026-06-05
Add CI workflow

JingsongLi:
## Summary
- Add GitHub Actions CI with two jobs:
  - **check**: cargo-deny license audit, `cargo fmt --check`, `cargo clippy -D warnings`
  - **rust-test**: build all targets + test with llvm-cov coverage summary
- Triggers on push/PR to main and release-* branches
- Concurrency control to cancel stale runs

## Test plan
- [ ] CI passes on this PR itself once core code is merged

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,80 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+      - release-*
+  pull_request:
+    branches:
+      - main
+      - release-*
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install cargo-deny
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-deny@0.19.0
+
+      - name: Check dependency licenses (Apache-compatible)
+        run: cargo deny check licenses
+
+      - name: Format
+        run: cargo fmt --all -- --check
+
+      - name: Clippy
+        run: cargo clippy --all-targets --workspace -- -D warnings
+
+  rust-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Rust Cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-
+
+      - name: Build
+        run: cargo build --all-targets
+
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+
+      - name: Test with coverage
+        run: cargo llvm-cov --summary-only
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,0 +1,22 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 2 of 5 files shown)


leaves12138 (changes requested):
Thanks for adding the CI workflow. I found a blocker in the current PR shape: this PR only adds `.github/workflows/ci.yml`, but its base branch does not contain a Cargo workspace yet.

On this PR head alone (`bc754d73dac916caf704209fd8b0f603290f7442`), even a minimal Cargo command fails:

`cargo metadata --no-deps --format-version 1`

with:

`error: could not find Cargo.toml in /tmp/paimon-vector-index-review or any parent directory`

So if this PR is merged before the core-foundation PR, the new CI workflow will run `cargo deny`, `cargo fmt`, `cargo clippy`, `cargo build`, and `cargo llvm-cov` on a repository that has no `Cargo.toml`, causing every push/PR check on `main` to fail.

Please either make this PR target/stack on top of the core-foundation branch, include the workspace files here, or merge it only after the Cargo workspace exists on `main`.

JingsongLi:
Thanks for the review. This branch is now rebased on top of `add-core-foundation` (PR #1), so it includes the Cargo workspace. Please merge PR #1 first, then this PR will cleanly apply on main with only the CI workflow diff.

leaves12138 (changes requested):
Thanks for updating this PR. The previous structural blocker is fixed now: this branch includes the Cargo workspace, so the workflow no longer runs Cargo commands in a repository without `Cargo.toml`.

I checked the stacked branch locally:

`cargo metadata --no-deps --format-version 1` succeeds and finds `paimon-vindex-core`.

`cargo test --all --all-features` also passes with 15 unit tests.

There is still one inherited blocker from the stacked core changes: `cargo fmt --all -- --check` currently fails with formatting diffs in `core/src/blas.rs`, `core/src/distance.rs`, `core/src/kmeans.rs`, and `core/src/pq.rs`. Because this workflow enforces that command, CI will fail until the core branch is formatted.

Once PR #1 is formatted and this PR is rebased/updated on top of it, I can re-check and approve the CI workflow.

leaves12138 (changes requested):
Thanks for the update. The stacked branch now has a Cargo workspace and the formatter issue is fixed:

`cargo fmt --all -- --check` passes.

`cargo test --all --all-features` also passes locally with 15 tests.

However, this workflow currently enforces:

`cargo clippy --all-targets --workspace -- -D warnings`

and that command still fails on the stacked core changes with multiple clippy warnings promoted to errors. So CI will fail until PR #1 either fixes or explicitly allows those lints.

Once PR #1 passes the exact clippy command and this PR is updated on top of it, I can re-check and approve the workflow.

leaves12138 (changes requested):
Thanks for cleaning this up. I re-checked the latest head (`afc1ed4df8b9e232055c582b582770bf4f62bb4a`), and this PR now only adds `.github/workflows/ci.yml`.

However, if the plan is to merge this PR before PR #1, this still has a blocker: current `main` does not have a Cargo workspace yet. After this workflow is merged, every push/PR against `main` will run commands such as:

`cargo deny check licenses`
`cargo fmt --all -- --check`
`cargo clippy --all-targets --workspace -- -D warnings`
`cargo build --all-targets`
`cargo llvm-cov --summary-only`

On the current base, even `cargo metadata --no-deps --format-version 1` fails with:

`error: could not find Cargo.toml in /tmp/paimon-vector-index-review or any parent directory`

So merging this PR first would make the repository CI fail until PR #1 lands.

If we want CI to land before the Rust workspace, please add a guard that skips the Cargo jobs when `Cargo.toml` is absent (or otherwise make the workflow a no-op until the workspace exists). If the workflow is intended to require a workspace unconditionally, then PR #1 should land first and this PR can be merged after that.

leaves12138 (changes requested):
Thanks for the update. I re-checked the latest head (`aabd7da5474382111b77163de5c6359fcc88eb36`): this PR now adds only `.github/workflows/ci.yml`, `Cargo.toml`, and `deny.toml`.

However, the added `Cargo.toml` is an empty virtual workspace:

```toml
[workspace]
members = []
resolver = "2"
```

This fixes `cargo metadata`, but it still does not make the CI commands pass. I verified locally on this PR head:

- `cargo metadata --no-deps --format-version 1` passes, but reports no workspace members.
- `cargo fmt --all -- --check` fails with `Failed to find targets`.
- `cargo clippy --all-targets --workspace -- -D warnings` fails with `The manifest is virtual, and the workspace has no members.`
- `cargo build --all-targets` fails with the same no-members error.

So merging this PR before adding any real crate would still make `main` CI fail.

Please either:

1. add a real minimal crate/member (for example `core/Cargo.toml` plus `core/src/lib.rs`) so these Cargo commands have a target, or
2. guard/skip the Cargo jobs while the workspace has no members.

After one of those is done, I can re-check and approve the CI PR.

leaves12138 (approved):
Thanks for the update. This version now adds a real minimal workspace member (`core`) together with the CI workflow, so the Cargo commands have an actual target even if this PR is merged before PR #1.

I verified locally on head `07fd9ffccce204d34f4bea11f023c20f92e89ce5`:

- `cargo metadata --no-deps --format-version 1` finds `paimon-vindex-core` as a workspace member.
- `cargo fmt --all -- --check` passes.
- `cargo clippy --all-targets --workspace -- -D warnings` passes.
- `cargo build --all-targets` passes.
- `cargo test --all --all-features` passes with 0 tests, as expected for the placeholder crate.

The placeholder crate has no dependencies, so the license check should be trivial once `cargo-deny` is installed by the workflow. `cargo-llvm-cov` is installed in the workflow as well; I could not run that locally because it is not installed in my environment, but the workspace/build/test path is now valid.

No further blocker from my side.

PR #3, opened 2026-06-05
Add opq, shuffler, and fastscan modules

JingsongLi:
## Summary
- **opq**: Optimized Product Quantization — learns a rotation matrix (SVD) to minimize quantization error before PQ encoding
- **shuffler**: Disk-based partition shuffler for large-scale IVF index building — write vectors with partition IDs, read back grouped by partition
- **fastscan**: Block-layout 4-bit PQ fast scan with SIMD (AVX2/NEON) — quantizes distance tables to u8 and accumulates via shuffle instructions

~975 lines of new code, all clippy-clean and tested (21 tests pass).

## Test plan
- [x] `cargo test --all` passes (21 tests)

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -22,6 +22,7 @@ edition = "2021"
 license = "Apache-2.0"
 
 [dependencies]
+nalgebra = "0.33"
 rand = "0.8"
 rayon = "1.10"
 matrixmultiply = "0.3"
--- a/core/src/fastscan.rs
+++ b/core/src/fastscan.rs
@@ -1,0 +1,479 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! FastScan: Faiss-style block layout (bbs=32) + vpshufb 32-way parallel lookup
+//! for 4-bit PQ codes.
+//!
+//! Block layout: 32 vectors per block, codes interleaved by sub-quantizer pair.
+//! Each block stores M/2 groups of 32 bytes (one byte per vector per sub-quant pair).
+//!
+//! Layout: [block0_pair0(32B), block0_pair1(32B), ..., block0_pairM/2(32B),
+//!          block1_pair0(32B), ...]
+
+/// Block size: 32 vectors per block (matches AVX2 register width).
+pub const BBS: usize = 32;
+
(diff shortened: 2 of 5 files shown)


leaves12138 (changes requested):
Thanks for adding these modules. I ran the existing test suite locally (`cargo fmt --all -- --check` and `cargo test --workspace --all-targets -- --nocapture`), and the checked-in tests pass. However, I found a correctness issue in the AVX2 FastScan path that should be fixed before merging.

`fastscan_blocks_avx2` stores the widened low/high halves directly into `q_vals`:

```rust
let pair_lo = _mm256_unpacklo_epi8(pair_dist, zero);
let pair_hi = _mm256_unpackhi_epi8(pair_dist, zero);
...
_mm256_storeu_si256(q_vals.as_mut_ptr() as *mut __m256i, accu_lo);
_mm256_storeu_si256(q_vals.as_mut_ptr().add(16) as *mut __m256i, accu_hi);
```

But `_mm256_unpacklo_epi8` / `_mm256_unpackhi_epi8` operate per 128-bit lane. For 32 input bytes `[0..31]`, the stored order becomes `[0..7, 16..23, 8..15, 24..31]`, not `[0..31]`. That means distances for each 32-vector block are written to the wrong vector positions on AVX2 machines.

A focused regression test that fails for me:

```rust
#[test]
fn test_fastscan_avx2_matches_row_major_after_aligned_start() {
    let m = 16;
    let cs = m / 2;
    let n = 257;
    let codes_row: Vec<u8> = (0..n * cs).map(|i| ((i * 13 + 7) % 256) as u8).collect();
    let sim_table: Vec<f32> = (0..m * 16).map(|i| (i as f32) * 0.05 + 1.0).collect();

    let packed = pack_codes_block_layout(&codes_row, n, cs);
    let mut got = vec![0.0f32; n];
    fastscan_4bit(&sim_table, &packed, n, m, &mut got);

    let mut expected = vec![0.0f32; n];
    crate::distance::scan_4bit_simd(&sim_table, &codes_row, n, m, &mut expected);

    // Start at 224 to skip the exact prefix and the alignment gap; this checks the SIMD block.
    for i in 224..n {
        assert!(
            (got[i] - expected[i]).abs() <= 1e-6,
            "first SIMD-block mismatch at {i}: got {}, expected {}",
            got[i],
            expected[i]
        );
    }
}
```

On my x86_64/AVX2 environment this fails with:

```text
first SIMD-block mismatch at 232: got 115.18274, expected 113.31137
```

Please either reorder the extracted `q_vals` before writing to `dists`, or change the AVX2 widening/extraction logic so that output positions preserve the original block order. It would also be good for `test_fastscan_large` to compare against a row-major/scalar reference for vectors beyond the first 200; currently it only checks that values are positive/reasonable, so this ordering bug is not caught by CI.

One smaller robustness note: `DiskShuffler::write_vector` also accepts any vector length and partition id, so a short vector can corrupt the record stream and an out-of-range partition can panic. Please consider validating `vector.len() == dim` and `partition_id < nlist` and returning `io::ErrorKind::InvalidInput`.

leaves12138 (approved):
Thanks for the quick update. I rechecked the latest head (`a8940b3203f4cf5510f430d7c8aeace5297693b7`).

The previous AVX2 block-order issue is fixed by reassembling the widened 128-bit lanes with `permute2x128`, and the large FastScan test now validates the quantized path against an exact reference beyond the first 200 vectors. The shuffler now also validates vector length and partition id with `InvalidInput`, with tests.

I reran:

- `cargo test --workspace --all-targets -- --nocapture`
- an extra temporary exact-reference FastScan regression (`n=257`, `m=16`), where the max error was about `0.05` within the configured 2% tolerance

GitHub checks are green too. Looks good to me now.

PR #5, opened 2026-06-07
Add IVF-PQ index module

JingsongLi:
## Summary
- **ivfpq**: IVF-PQ index — the central index structure tying together kmeans (coarse quantizer), PQ (compression), OPQ (rotation), and fastscan (4-bit SIMD acceleration)
- Train/add/search pipeline with rayon parallelization, optional ID filtering, max_codes budget, precomputed tables, and distributed build support (from_trained + merge_from)
- Supports L2, InnerProduct, and Cosine metrics; both 8-bit and 4-bit PQ

Reader-based search functions (`search_with_reader`, `search_batch_reader`) are deferred to a follow-up PR with the `io` module.

~1133 lines of new code, all clippy-clean and tested (30 tests pass, 7 new).

## Test plan
- [x] `cargo test --all` passes (30 tests, 7 new ivfpq tests)
  - test_build_and_search_l2: L2 distance build + search, verifies top-1 recall
  - test_build_and_search_ip: InnerProduct distance build + search
  - test_search_with_filter: ID-filtered search, verifies filter enforcement
  - test_batch_search: multi-query batch search, verifies self-recall
  - test_4bit_ivfpq: 4-bit PQ mode, verifies recall and ~50% storage reduction
  - test_max_codes_early_termination: bounded scan budget
  - test_from_trained_and_merge: distributed build workflow (train once, add separately, merge)

### DIFF
--- a/core/src/ivfpq.rs
+++ b/core/src/ivfpq.rs
@@ -141,30 +141,35 @@ impl IVFPQIndex {
             data[..n * d].to_vec()
         };
 
+        // When OPQ is enabled, jointly train rotation + PQ, then project data.
+        // IVF centroids must be trained on projected (rotated) data since
+        // add() and search() assign rotated vectors via preprocess_queries().
         let effective_data = if let Some(ref mut opq) = self.opq {
             opq.train(&train_data, n, &mut self.pq);
-            &train_data
+            let mut projected = vec![0.0f32; n * d];
+            opq.apply_batch(&train_data, &mut projected, n);
+            projected
         } else {
-            &train_data
+            train_data
         };
 
         let km_config = KMeansConfig::default();
         self.quantizer_centroids =
-            kmeans::kmeans_train(&km_config, effective_data, n, d, self.nlist);
+            kmeans::kmeans_train(&km_config, &effective_data, n, d, self.nlist);
 
-        if let Some(ref opq) = self.opq {
+        if self.opq.is_some() {
+            // OPQ already trained PQ on rotated data; for by_residual,
+            // retrain PQ on residuals so it matches what add/search compute.
             if self.by_residual {
-                let mut projected = vec![0.0f32; n * d];
-                opq.apply_batch(effective_data, &mut projected, n);
                 let residuals =
-                    compute_residuals(&projected, n, d, &self.quantizer_centroids, self.nlist);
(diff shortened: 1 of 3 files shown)


leaves12138 (changes requested):
Thanks for adding the IVF-PQ module. I ran `cargo fmt --all -- --check` and `cargo test --workspace --all-targets -- --nocapture`; the checked-in tests pass locally.

I found one stale search-structure issue that should be fixed before merging. `add_batch` invalidates `fastscan_codes` after appending vectors, but it does not invalidate `precomputed_table`. Since `search` and `search_batch` enable the precomputed path whenever `precomputed_table` is non-empty, the sequence `add(); build_precomputed_table(); add(); search()` keeps using a table built before the second add. `merge_from` already clears both structures, so `add` should do the same.

Minimal regression:

```rust
#[test]
fn test_precomputed_table_invalidated_after_add() {
    let d = 16;
    let nlist = 4;
    let m = 4;
    let n = 500;

    let data = generate_clustered_data(n * 2, d, 4, 42);
    let ids_a: Vec<i64> = (0..n as i64).collect();
    let ids_b: Vec<i64> = (n as i64..2 * n as i64).collect();

    index.train(&data[..n * d], n);
    index.add(&data[..n * d], &ids_a, n);

    index.build_precomputed_table();
    assert!(!index.precomputed_table.is_empty());

    index.add(&data[n * d..], &ids_b, n);
    assert!(
        index.precomputed_table.is_empty(),
        "precomputed_table must be cleared after add()"
    );
}
```

Please clear `self.precomputed_table` in `add_batch` when new vectors are appended, and add a regression test analogous to `test_fastscan_invalidated_after_add`.

Minor robustness note: `k = 0` currently panics in `TopKHeap::push`; it may be worth handling or documenting separately, but the stale precomputed table is the blocker here.

leaves12138 (approved):
Thanks for the update. I re-reviewed the latest head and verified that the stale `precomputed_table` path is now invalidated from `add_batch`, with a regression test added. `k = 0` is also handled in `TopKHeap`.

I reran:

- `cargo test --workspace --all-targets -- --nocapture`

Both pass locally, and the CI checks are green. The IVF-PQ add/search/precomputed/fastscan invalidation paths look good to me now.

PR #6, opened 2026-06-08
Add io module and reader-based search

JingsongLi:
## Summary
- **io**: Binary serialization format for IVF-PQ indexes — delta-varint ID compression (typically 3-5x smaller than raw int64) + transposed code layout for cache-friendly SIMD scan
- **Reader-based search**: `search_with_reader`, `search_with_reader_filter`, `search_batch_reader` — lazy-loading reader (header-only open), reads inverted lists on demand, parallel per-list scanning
- **Batch search optimization**: `search_batch_reader` groups queries by probed list, reading each unique list once instead of nq*nprobe I/O ops

~1260 lines of new code, all clippy-clean and tested (43 tests pass, 7 new).

## Test plan
- [x] `cargo test --all` passes (43 tests, 7 new)
  - io: test_varint_roundtrip, test_delta_varint_ids_roundtrip
  - io: test_write_read_roundtrip_delta_ids (sorted IDs, delta encoding)
  - io: test_space_savings (100K vectors, verifies >10% compression)
  - ivfpq: test_write_read_search (serialize → deserialize → search)
  - ivfpq: test_write_read_search_with_filter (reader + ID filter)
  - ivfpq: test_big_batch_search (batch reader, 20 queries)

### DIFF
--- a/core/src/io.rs
+++ b/core/src/io.rs
@@ -109,12 +109,19 @@ fn decode_varint(buf: &[u8], pos: &mut usize) -> io::Result<u64> {
         }
         let b = buf[*pos] as u64;
         *pos += 1;
-        val |= (b & 0x7F) << shift;
+        let payload = b & 0x7F;
+        if shift == 63 && payload > 1 {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidData,
+                "varint exceeds u64 range",
+            ));
+        }
+        val |= payload << shift;
         if b & 0x80 == 0 {
             break;
         }
         shift += 7;
-        if shift >= 64 {
+        if shift > 63 {
             return Err(io::Error::new(
                 io::ErrorKind::InvalidData,
                 "varint exceeds 64 bits",
@@ -755,6 +762,14 @@ mod tests {
         assert_eq!(decode_varint(&buf, &mut pos).unwrap(), 128);
         assert_eq!(decode_varint(&buf, &mut pos).unwrap(), 16383);
         assert_eq!(decode_varint(&buf, &mut pos).unwrap(), 1_000_000);
+    }
+
+    #[test]
+    fn test_varint_above_u64_max_returns_error() {
+        let mut bytes = vec![0xFFu8; 9];
+        bytes.push(0x02); // 10th byte with payload > 1 at shift=63
+        let mut pos = 0;
+        assert!(decode_varint(&bytes, &mut pos).is_err());
     }
 
     #[test]
--- a/core/src/io.rs
+++ b/core/src/io.rs
@@ -125,7 +125,7 @@ fn decode_varint(buf: &[u8], pos: &mut usize) -> io::Result<u64> {
 }
 
 /// Encode sorted i64 IDs as delta-varint. Returns (base_id, encoded_bytes).
-/// IDs must be sorted ascending; panics if any id < prev.
+/// Uses unsigned subtraction to handle the full i64 range without overflow.
 fn encode_delta_varint_ids(ids: &[i64]) -> (i64, Vec<u8>) {
     if ids.is_empty() {
         return (0, Vec::new());
@@ -134,26 +134,23 @@ fn encode_delta_varint_ids(ids: &[i64]) -> (i64, Vec<u8>) {
     let mut buf = Vec::with_capacity(ids.len() * 2);
     let mut prev = base;
     for &id in ids {
-        let delta = id.checked_sub(prev).expect("IDs must be sorted ascending") as u64;
+        let delta = (id as u64).wrapping_sub(prev as u64);
         encode_varint(delta, &mut buf);
         prev = id;
     }
     (base, buf)
 }
 
-/// Decode delta-varint encoded IDs.
+/// Decode delta-varint encoded IDs using wrapping unsigned arithmetic
+/// (inverse of encode_delta_varint_ids).
 fn decode_delta_varint_ids(base: i64, buf: &[u8], count: usize) -> io::Result<Vec<i64>> {
     let mut ids = Vec::with_capacity(count);
     let mut pos = 0;
-    let mut current = base;
+    let mut current = base as u64;
     for _ in 0..count {
         let delta = decode_varint(buf, &mut pos)?;
-        let delta_i64 = i64::try_from(delta)
-            .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "delta exceeds i64 range"))?;
-        current = current.checked_add(delta_i64).ok_or_else(|| {
-            io::Error::new(io::ErrorKind::InvalidData, "reconstructed ID overflows i64")
-        })?;
-        ids.push(current);
+        current = current.wrapping_add(delta);
+        ids.push(current as i64);
     }
     Ok(ids)
 }
@@ -571,9 +568,16 @@ impl<R: SeekRead> IVFPQIndexReader<R> {
         self.reader.seek(offset)?;
 
         let ids = if self.delta_ids {
-            // Delta-varint format: [base_id: i64][id_bytes_len: u32][id_bytes...]
+            // Delta-varint format: [base_id: i64][id_bytes_len: i32][id_bytes...]
             let base_id = read_i64_le(&mut self.reader)?;
-            let id_bytes_len = read_i32_le(&mut self.reader)? as usize;
+            let id_bytes_len_raw = read_i32_le(&mut self.reader)?;
+            if id_bytes_len_raw < 0 {
+                return Err(io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    "negative id_bytes_len",
+                ));
+            }
+            let id_bytes_len = id_bytes_len_raw as usize;
             let mut id_bytes = vec![0u8; id_bytes_len];
             self.reader.read_exact(&mut id_bytes)?;
             decode_delta_varint_ids(base_id, &id_bytes, count)?
@@ -885,7 +889,7 @@ mod tests {
     }
 
     #[test]
-    fn test_corrupt_delta_ids_overflow_returns_error() {
+    fn test_negative_id_bytes_len_returns_error() {
         let mut buf = Vec::new();
         buf.extend_from_slice(&MAGIC.to_le_bytes());
         buf.extend_from_slice(&VERSION.to_le_bytes());
@@ -908,18 +912,23 @@ mod tests {
         buf.extend_from_slice(&1i32.to_le_bytes()); // count=1
         buf.extend_from_slice(&0i32.to_le_bytes()); // padding
 
-        // base_id = i64::MAX, delta = 1 → overflow
-        buf.extend_from_slice(&i64::MAX.to_le_bytes());
-        buf.extend_from_slice(&1i32.to_le_bytes()); // id_bytes_len = 1
-        buf.push(1u8); // varint delta = 1
-        buf.extend_from_slice(&[0u8; 1]); // one PQ code byte
+        buf.extend_from_slice(&0i64.to_le_bytes()); // base_id
+        buf.extend_from_slice(&(-1i32).to_le_bytes()); // negative id_bytes_len
 
         let mut cursor = Cursor::new(&buf);
         let mut reader = IVFPQIndexReader::open(&mut cursor).unwrap();
         let result = reader.read_inverted_list(0);
         assert!(
             result.is_err(),
-            "overflowed delta IDs should return error, not panic"
+            "negative id_bytes_len should return error, not panic"
         );
+    }
+
+    #[test]
+    fn test_large_gap_ids_roundtrip() {
+        let ids = vec![i64::MIN, 0, i64::MAX];
+        let (base, encoded) = encode_delta_varint_ids(&ids);
+        let decoded = decode_delta_varint_ids(base, &encoded, ids.len()).unwrap();
+        assert_eq!(decoded, ids);
     }
 }
--- a/core/src/io.rs
+++ b/core/src/io.rs
@@ -1,0 +1,797 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::distance::MetricType;
+use crate::ivfpq::IVFPQIndex;
+use crate::opq::OPQMatrix;
+use crate::pq::ProductQuantizer;
+use std::io;
+
+pub const MAGIC: u32 = 0x49565051; // "IVPQ"
+pub const VERSION: u32 = 1;
+pub const HEADER_SIZE: usize = 64;
+
+pub const FLAG_HAS_OPQ: u32 = 1 << 0;
+pub const FLAG_BY_RESIDUAL: u32 = 1 << 1;
+pub const FLAG_DELTA_IDS: u32 = 1 << 2;
+pub const FLAG_TRANSPOSED_CODES: u32 = 1 << 3;
+
+pub trait SeekRead: Send {
+    fn seek(&mut self, pos: u64) -> io::Result<()>;
+    fn read_exact(&mut self, buf: &mut [u8]) -> io::Result<()>;
+
+    /// Positional read: read `buf.len()` bytes at `pos` without changing the cursor.
+    /// Thread-safe if the underlying implementation supports it (e.g., pread(2)).
+    /// Default implementation falls back to seek + read_exact.
+    fn pread(&mut self, pos: u64, buf: &mut [u8]) -> io::Result<()> {
+        self.seek(pos)?;
+        self.read_exact(buf)
+    }
+
+    /// Whether this implementation supports true concurrent pread (no shared cursor).
+    fn supports_concurrent_pread(&self) -> bool {
+        false
+    }
+}
+
+pub trait SeekWrite: Send {
+    fn write_all(&mut self, buf: &[u8]) -> io::Result<()>;
+    fn pos(&self) -> u64;
+}
+
+impl<T: io::Read + io::Seek + Send> SeekRead for T {
+    fn seek(&mut self, pos: u64) -> io::Result<()> {
+        io::Seek::seek(self, io::SeekFrom::Start(pos))?;
+        Ok(())
+    }
+
+    fn read_exact(&mut self, buf: &mut [u8]) -> io::Result<()> {
+        io::Read::read_exact(self, buf)
+    }
+}
+
+pub struct PosWriter<W: io::Write> {
+    inner: W,
+    pos: u64,
+}
+
+impl<W: io::Write> PosWriter<W> {
+    pub fn new(inner: W) -> Self {
+        PosWriter { inner, pos: 0 }
+    }
+}
+
+impl<W: io::Write + Send> SeekWrite for PosWriter<W> {
+    fn write_all(&mut self, buf: &[u8]) -> io::Result<()> {
+        self.inner.write_all(buf)?;
+        self.pos += buf.len() as u64;
+        Ok(())
+    }
+
+    fn pos(&self) -> u64 {
+        self.pos
+    }
+}
+
+// --- Varint encoding ---
+
+fn encode_varint(mut val: u64, buf: &mut Vec<u8>) {
+    while val >= 0x80 {
+        buf.push((val as u8) | 0x80);
+        val >>= 7;
+    }
+    buf.push(val as u8);
+}
+
+fn decode_varint(buf: &[u8], pos: &mut usize) -> u64 {
+    let mut val: u64 = 0;
+    let mut shift = 0;
+    loop {
+        let b = buf[*pos] as u64;
+        *pos += 1;
+        val |= (b & 0x7F) << shift;
+        if b & 0x80 == 0 {
+            break;
+        }
+        shift += 7;
+    }
+    val
+}
+
+/// Encode sorted i64 IDs as delta-varint. Returns (base_id, encoded_bytes).
+fn encode_delta_varint_ids(ids: &[i64]) -> (i64, Vec<u8>) {
+    if ids.is_empty() {
+        return (0, Vec::new());
+    }
+    let base = ids[0];
+    let mut buf = Vec::with_capacity(ids.len() * 2);
+    let mut prev = base;
+    for &id in ids {
+        let delta = (id - prev) as u64;
+        encode_varint(delta, &mut buf);
+        prev = id;
+    }
+    (base, buf)
+}
+
+/// Decode delta-varint encoded IDs.
+fn decode_delta_varint_ids(base: i64, buf: &[u8], count: usize) -> Vec<i64> {
+    let mut ids = Vec::with_capacity(count);
+    let mut pos = 0;
+    let mut current = base;
+    for _ in 0..count {
+        let delta = decode_varint(buf, &mut pos) as i64;
+        current += delta;
+        ids.push(current);
+    }
+    ids
+}
+
+// --- Read/write helpers ---
+
+fn write_u32_le(out: &mut dyn SeekWrite, v: u32) -> io::Result<()> {
+    out.write_all(&v.to_le_bytes())
+}
+
+fn write_i32_le(out: &mut dyn SeekWrite, v: i32) -> io::Result<()> {
+    out.write_all(&v.to_le_bytes())
+}
+
+fn write_i64_le(out: &mut dyn SeekWrite, v: i64) -> io::Result<()> {
+    out.write_all(&v.to_le_bytes())
+}
+
+fn write_f32_slice(out: &mut dyn SeekWrite, data: &[f32]) -> io::Result<()> {
+    let bytes: Vec<u8> = data.iter().flat_map(|f| f.to_le_bytes()).collect();
+    out.write_all(&bytes)
+}
+
+fn read_u32_le(reader: &mut dyn SeekRead) -> io::Result<u32> {
+    let mut buf = [0u8; 4];
+    reader.read_exact(&mut buf)?;
+    Ok(u32::from_le_bytes(buf))
+}
+
+fn read_i32_le(reader: &mut dyn SeekRead) -> io::Result<i32> {
+    let mut buf = [0u8; 4];
+    reader.read_exact(&mut buf)?;
+    Ok(i32::from_le_bytes(buf))
+}
+
+fn read_i64_le(reader: &mut dyn SeekRead) -> io::Result<i64> {
+    let mut buf = [0u8; 8];
+    reader.read_exact(&mut buf)?;
+    Ok(i64::from_le_bytes(buf))
+}
+
+fn read_f32_vec(reader: &mut dyn SeekRead, count: usize) -> io::Result<Vec<f32>> {
+    let mut buf = vec![0u8; count * 4];
+    reader.read_exact(&mut buf)?;
+    let floats: Vec<f32> = buf
+        .chunks_exact(4)
+        .map(|c| f32::from_le_bytes([c[0], c[1], c[2], c[3]]))
+        .collect();
+    Ok(floats)
+}
+
+/// Write a complete IVF-PQ index with delta-varint ID encoding.
+pub fn write_index(index: &IVFPQIndex, out: &mut dyn SeekWrite) -> io::Result<()> {
+    let d = index.d;
+    let nlist = index.nlist;
+    let m = index.pq.m;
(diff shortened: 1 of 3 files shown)


leaves12138 (changes requested):
Thanks for the PR. I reviewed latest head `f1b6c91e8029960656f068f0d5a9731f142d6535`.

What I verified locally on a clean checkout:

All of those pass. I also added temporary local regressions comparing reader search with in-memory search, checking batch-reader results against repeated single-reader searches, and checking raw-ID 4-bit roundtrip search; those passed too.

I found one remaining correctness/robustness blocker in the new delta-varint reader path. `decode_varint` indexes `buf[*pos]` without checking that `pos < buf.len()`, and `decode_delta_varint_ids` calls it `count` times without validating that the encoded bytes actually contain enough varints. Therefore a corrupted or truncated index file can panic instead of returning an `io::Error` from `read_inverted_list`.

The problematic path is:

```rust
fn decode_varint(buf: &[u8], pos: &mut usize) -> u64 {
    let mut val: u64 = 0;
    let mut shift = 0;
    loop {
        let b = buf[*pos] as u64; // panics when id_bytes is truncated
        *pos += 1;
        ...
    }
}
```

Minimal regression:

```rust
#[test]
fn corrupt_delta_ids_should_return_error_not_panic() {
    let mut buf = Vec::new();
    fn push_i32(buf: &mut Vec<u8>, v: i32) { buf.extend_from_slice(&v.to_le_bytes()); }
    fn push_i64(buf: &mut Vec<u8>, v: i64) { buf.extend_from_slice(&v.to_le_bytes()); }
    fn push_u32(buf: &mut Vec<u8>, v: u32) { buf.extend_from_slice(&v.to_le_bytes()); }

    push_u32(&mut buf, MAGIC);
    push_u32(&mut buf, VERSION);
    push_i32(&mut buf, 4);   // d
    push_i32(&mut buf, 1);   // nlist
    push_i32(&mut buf, 1);   // m
    push_i32(&mut buf, 256); // ksub
    push_i32(&mut buf, 4);   // dsub
    push_u32(&mut buf, MetricType::L2 as u32);
    push_i64(&mut buf, 1);
    push_u32(&mut buf, FLAG_DELTA_IDS | FLAG_TRANSPOSED_CODES | FLAG_BY_RESIDUAL);
    buf.extend_from_slice(&[0u8; 20]);
    buf.extend_from_slice(&vec![0u8; 4 * 4]);           // quantizer centroids
    buf.extend_from_slice(&vec![0u8; 1 * 256 * 4 * 4]); // pq centroids

    let list_offset = buf.len() as i64 + 16;
    push_i64(&mut buf, list_offset);
    push_i32(&mut buf, 1); // one id expected
    push_i32(&mut buf, 0);

    push_i64(&mut buf, 123); // base id
    push_i32(&mut buf, 0);   // id_bytes_len too short for count=1

    let mut cursor = std::io::Cursor::new(&buf);
    let mut reader = IVFPQIndexReader::open(&mut cursor).unwrap();
    assert!(reader.read_inverted_list(0).is_err());
}
```

On the current PR this panics with `index out of bounds: the len is 0 but the index is 0` at `core/src/io.rs:104`.

Please make varint decoding fallible, e.g. return `io::Result<u64>` / `Result<u64, String>`, detect unterminated varints and excessive shifts, and have `read_inverted_list` surface malformed `id_bytes_len` / truncated delta-id data as `io::ErrorKind::InvalidData` instead of panicking. It would also be good to add a corrupt-file regression test for this path.

leaves12138 (changes requested):
Thanks for the update. I re-reviewed latest head `05e7b05e7b9e35b4731177455edd63d088bc7619`.

The previous truncated-varint panic is fixed: `decode_varint` is now fallible and the new corrupt-file regression covers the zero-length `id_bytes` case.

I found one remaining robustness blocker in the same delta-ID decode path. `decode_delta_varint_ids` still performs unchecked signed accumulation:

```rust
let delta = decode_varint(buf, &mut pos)? as i64;
current += delta;
```

A malformed index can encode a delta that overflows the current `i64` ID. In debug builds this still panics instead of returning `io::Error`; in release builds it may wrap and return corrupted IDs. For example, a list with `base_id = i64::MAX` and one encoded delta byte `0x01` reaches `current += delta` and panics with `attempt to add with overflow`.

Minimal local regression:

```rust
#[test]
fn corrupt_delta_ids_overflow_should_not_panic() {
    let mut buf = Vec::new();
    buf.extend_from_slice(&MAGIC.to_le_bytes());
    buf.extend_from_slice(&VERSION.to_le_bytes());
    buf.extend_from_slice(&4i32.to_le_bytes()); // d
    buf.extend_from_slice(&1i32.to_le_bytes()); // nlist
    buf.extend_from_slice(&1i32.to_le_bytes()); // m
    buf.extend_from_slice(&256i32.to_le_bytes()); // ksub
    buf.extend_from_slice(&4i32.to_le_bytes()); // dsub
    buf.extend_from_slice(&(MetricType::L2 as u32).to_le_bytes());
    buf.extend_from_slice(&1i64.to_le_bytes()); // total_vectors
    let flags = FLAG_DELTA_IDS | FLAG_TRANSPOSED_CODES | FLAG_BY_RESIDUAL;
    buf.extend_from_slice(&flags.to_le_bytes());
    buf.extend_from_slice(&[0u8; 20]);

    buf.extend_from_slice(&[0u8; 16]);
    buf.extend_from_slice(&vec![0u8; 256 * 4 * 4]);

    let list_data_offset = buf.len() as i64 + 16;
    buf.extend_from_slice(&list_data_offset.to_le_bytes());
    buf.extend_from_slice(&1i32.to_le_bytes()); // count=1
    buf.extend_from_slice(&0i32.to_le_bytes());

    buf.extend_from_slice(&i64::MAX.to_le_bytes()); // base_id
    buf.extend_from_slice(&1i32.to_le_bytes()); // id_bytes_len
    buf.push(1u8); // delta = 1, overflows i64::MAX + 1
    buf.extend_from_slice(&[0u8; 1]); // one 8-bit PQ code for m=1

    let mut cursor = std::io::Cursor::new(&buf);
    let mut reader = IVFPQIndexReader::open(&mut cursor).unwrap();
    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
        reader.read_inverted_list(0)
    }));
    assert!(result.is_ok(), "read_inverted_list should return an error, not panic");
    assert!(result.unwrap().is_err(), "overflowed delta IDs should be rejected");
}
```

Please make delta reconstruction checked as well, e.g. accumulate in `i128` or use checked arithmetic and return `io::ErrorKind::InvalidData` when the reconstructed ID is outside the `i64` range. It would also be good to consider the writer side: `encode_delta_varint_ids` currently uses `id - prev` on `i64`, which can also overflow for otherwise sorted `i64` IDs with a very large gap.

Validation I ran before the extra local regression:

Those pass on the PR, but the overflow regression above still fails locally.

leaves12138 (changes requested):
Thanks for the quick update. I re-reviewed latest head `51584245e99dcbaa9570ca9326c8d5a2b5a83a8b`.

The previous decode-side overflow case is fixed and covered by `test_corrupt_delta_ids_overflow_returns_error`. I also re-ran:

Those pass on the PR.

However, there are still two robustness issues in the same IO path.

1. The writer-side delta encoding still panics for valid sorted `i64` IDs with a large gap.

`encode_delta_varint_ids` currently does:

```rust
let delta = id.checked_sub(prev).expect("IDs must be sorted ascending") as u64;
```

This treats arithmetic overflow the same as unsorted input. For sorted IDs like `[i64::MIN, i64::MAX]`, `id >= prev` is true, but `i64::MAX - i64::MIN` cannot fit in `i64`, so `write_index` panics. The delta is naturally non-negative and fits in `u64`, so this should be computed via a wider integer, e.g. `(id as i128 - prev as i128)` and then checked/cast to `u64`, or the writer should return `io::Error` instead of panicking.

Minimal local regression:

```rust
#[test]
fn extreme_sorted_i64_ids_should_roundtrip() {
    let d = 4;
    let nlist = 1;
    let m = 1;
    let n = 2;
    let data = vec![0.0f32, 0.0, 0.0, 0.0, 1.0, 1.0, 1.0, 1.0];
    let ids = vec![i64::MIN, i64::MAX];

    index.train(&data, n);
    index.add(&data, &ids, n);

    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
        let mut buf = Vec::new();
        let mut writer = PosWriter::new(&mut buf);
        write_index(&index, &mut writer).unwrap();
        let mut cursor = std::io::Cursor::new(&buf);
        let mut reader = IVFPQIndexReader::open(&mut cursor).unwrap();
        reader.read_inverted_list(0).unwrap().0
    }));
    assert!(result.is_ok(), "writer/reader should not panic for sorted i64 IDs");
    assert_eq!(result.unwrap(), ids);
}
```

2. Negative `id_bytes_len` in a malformed file still panics before returning `io::Error`.

`read_inverted_list` reads the length as `i32` and casts directly to `usize`:

```rust
let id_bytes_len = read_i32_le(&mut self.reader)? as usize;
let mut id_bytes = vec![0u8; id_bytes_len];
```

If the encoded length is `-1`, this becomes a huge `usize` and panics with `capacity overflow`. The reader should reject negative lengths with `io::ErrorKind::InvalidData` before allocating. It would also be safer to validate list counts and offsets similarly when loading the offset table.

Minimal local regression:

```rust
#[test]
fn negative_delta_id_bytes_len_should_return_error_not_panic() {
    let mut buf = Vec::new();
    buf.extend_from_slice(&MAGIC.to_le_bytes());
    buf.extend_from_slice(&VERSION.to_le_bytes());
    buf.extend_from_slice(&4i32.to_le_bytes()); // d
    buf.extend_from_slice(&1i32.to_le_bytes()); // nlist
    buf.extend_from_slice(&1i32.to_le_bytes()); // m
    buf.extend_from_slice(&256i32.to_le_bytes()); // ksub
    buf.extend_from_slice(&4i32.to_le_bytes()); // dsub
    buf.extend_from_slice(&(MetricType::L2 as u32).to_le_bytes());
    buf.extend_from_slice(&1i64.to_le_bytes()); // total_vectors
    let flags = FLAG_DELTA_IDS | FLAG_TRANSPOSED_CODES | FLAG_BY_RESIDUAL;
    buf.extend_from_slice(&flags.to_le_bytes());
    buf.extend_from_slice(&[0u8; 20]);
    buf.extend_from_slice(&[0u8; 16]);
    buf.extend_from_slice(&vec![0u8; 256 * 4 * 4]);

    let list_data_offset = buf.len() as i64 + 16;
    buf.extend_from_slice(&list_data_offset.to_le_bytes());
    buf.extend_from_slice(&1i32.to_le_bytes()); // count=1
    buf.extend_from_slice(&0i32.to_le_bytes());
    buf.extend_from_slice(&0i64.to_le_bytes());
    buf.extend_from_slice(&(-1i32).to_le_bytes()); // invalid id_bytes_len

    let mut cursor = std::io::Cursor::new(&buf);
    let mut reader = IVFPQIndexReader::open(&mut cursor).unwrap();
    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
        reader.read_inverted_list(0)
    }));
    assert!(result.is_ok(), "negative id_bytes_len should return error, not panic");
    assert!(result.unwrap().is_err());
}
```

Please make these two paths return `io::Error` instead of panicking, and add regressions for them.

leaves12138 (changes requested):
Thanks for the update. I re-reviewed latest head `7884e2bee87212015f1f84f1ec4fff138933616f`.

The two previously reported cases are partially addressed:
- Large-gap sorted IDs such as `[i64::MIN, 0, i64::MAX]` now round-trip in the helper test.
- Negative `id_bytes_len` is rejected before allocation.
- `cargo fmt --all -- --check`, `cargo clippy --all-targets --workspace -- -D warnings`, and `cargo test --all -- --nocapture` pass locally.

I still found two blockers in the IO reader robustness/correctness path.

1. Invalid delta streams can now be silently decoded into wrong wrapped IDs.

The new decoder uses wrapping unsigned accumulation:

```rust
let mut current = base as u64;
for _ in 0..count {
    let delta = decode_varint(buf, &mut pos)?;
    current = current.wrapping_add(delta);
    ids.push(current as i64);
}
```

This makes the writer-side full-range roundtrip work, but it also accepts malformed on-disk data where the reconstructed signed ID order wraps around. For example, `base_id = i64::MAX` and `delta = 1` is decoded as `i64::MIN` instead of returning `InvalidData`. That is worse than the previous panic: a corrupt file can return a wrong ID silently.

Minimal regression:

```rust
#[test]
fn delta_ids_wraparound_should_return_error_not_wrong_ids() {
    let mut buf = Vec::new();
    buf.extend_from_slice(&MAGIC.to_le_bytes());
    buf.extend_from_slice(&VERSION.to_le_bytes());
    buf.extend_from_slice(&4i32.to_le_bytes()); // d
    buf.extend_from_slice(&1i32.to_le_bytes()); // nlist
    buf.extend_from_slice(&1i32.to_le_bytes()); // m
    buf.extend_from_slice(&256i32.to_le_bytes()); // ksub
    buf.extend_from_slice(&4i32.to_le_bytes()); // dsub
    buf.extend_from_slice(&(MetricType::L2 as u32).to_le_bytes());
    buf.extend_from_slice(&1i64.to_le_bytes()); // total_vectors
    let flags = FLAG_DELTA_IDS | FLAG_TRANSPOSED_CODES | FLAG_BY_RESIDUAL;
    buf.extend_from_slice(&flags.to_le_bytes());
    buf.extend_from_slice(&[0u8; 20]);
    buf.extend_from_slice(&[0u8; 16]);
    buf.extend_from_slice(&vec![0u8; 256 * 4 * 4]);

    let list_data_offset = buf.len() as i64 + 16;
    buf.extend_from_slice(&list_data_offset.to_le_bytes());
    buf.extend_from_slice(&1i32.to_le_bytes()); // count=1
    buf.extend_from_slice(&0i32.to_le_bytes());

    buf.extend_from_slice(&i64::MAX.to_le_bytes()); // base_id
    buf.extend_from_slice(&1i32.to_le_bytes()); // id_bytes_len
    buf.push(1u8); // wraps to i64::MIN with the current decoder
    buf.extend_from_slice(&[0u8; 1]); // one PQ code byte

    let mut cursor = std::io::Cursor::new(&buf);
    let mut reader = IVFPQIndexReader::open(&mut cursor).unwrap();
    let result = reader.read_inverted_list(0);
    assert!(result.is_err(), "wrapped delta IDs should be rejected, not decoded as wrong IDs");
}
```

If the format wants to support the full signed `i64` range while preserving sorted-ID semantics, the decoder should validate reconstructed IDs remain monotonically non-decreasing in signed order. The writer can compute deltas with `i128`, but the reader should reject any decoded value that would go backwards as an `i64` sequence.

2. Negative list counts are still accepted and can panic later.

`ensure_loaded` stores `list_counts[i] = read_i32_le(...)` without validating it. Then `read_inverted_list` does:

```rust
let count = self.list_counts[list_id] as usize;
```

For a malformed offset-table entry with `count = -1`, this becomes a huge `usize`. If the file also contains enough bytes to get past the ID length read, the code tries to allocate a huge codes buffer and panics with `capacity overflow`.

Minimal regression:

```rust
#[test]
fn negative_list_count_with_payload_should_return_error_not_panic() {
    let mut buf = Vec::new();
    buf.extend_from_slice(&MAGIC.to_le_bytes());
    buf.extend_from_slice(&VERSION.to_le_bytes());
    buf.extend_from_slice(&4i32.to_le_bytes()); // d
    buf.extend_from_slice(&1i32.to_le_bytes()); // nlist
    buf.extend_from_slice(&1i32.to_le_bytes()); // m
    buf.extend_from_slice(&256i32.to_le_bytes()); // ksub
    buf.extend_from_slice(&4i32.to_le_bytes()); // dsub
    buf.extend_from_slice(&(MetricType::L2 as u32).to_le_bytes());
    buf.extend_from_slice(&1i64.to_le_bytes()); // total_vectors
    let flags = FLAG_DELTA_IDS | FLAG_TRANSPOSED_CODES | FLAG_BY_RESIDUAL;
    buf.extend_from_slice(&flags.to_le_bytes());
    buf.extend_from_slice(&[0u8; 20]);
    buf.extend_from_slice(&[0u8; 16]);
    buf.extend_from_slice(&vec![0u8; 256 * 4 * 4]);

    let list_data_offset = buf.len() as i64 + 16;
    buf.extend_from_slice(&list_data_offset.to_le_bytes());
    buf.extend_from_slice(&(-1i32).to_le_bytes()); // invalid count
    buf.extend_from_slice(&0i32.to_le_bytes());
    buf.extend_from_slice(&0i64.to_le_bytes()); // base_id
    buf.extend_from_slice(&0i32.to_le_bytes()); // id_bytes_len

    let mut cursor = std::io::Cursor::new(&buf);
    let mut reader = IVFPQIndexReader::open(&mut cursor).unwrap();
    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
        reader.read_inverted_list(0)
    }));
    assert!(result.is_ok(), "negative list count should return error, not panic");
    assert!(result.unwrap().is_err());
}
```

Please validate offset-table counts (and ideally offsets) as malformed input before any `as usize` cast or allocation, and add regressions for both cases.

leaves12138 (changes requested):
Thanks for the update. I re-reviewed latest head `813ed79a3e7aa72041514a60e39b9c94eed62727`.

The two blockers from the previous round are fixed:
- Wrapped delta streams are now rejected as non-monotonic.
- Negative list counts are rejected during `ensure_loaded`.

Validation passed locally:
- Focused regressions for `test_delta_ids_wraparound_returns_error`, `test_negative_list_count_returns_error`, `test_large_gap_ids_roundtrip`, and `write_read` tests.

I found one remaining robustness blocker in the same malformed-file class: header dimensions are still read as `i32` and cast directly to `usize` before validation. A corrupt header with negative `d` or negative `nlist` can panic later during allocation/size multiplication in `ensure_loaded` instead of returning `io::Error`.

Examples from local regressions:

- `d = -1` reaches `read_f32_vec(&mut self.reader, nlist * d)` and panics with `attempt to multiply with overflow`.
- `nlist = -1` reaches `read_f32_vec(&mut self.reader, nlist * d)` / related section-size calculation and also panics with `attempt to multiply with overflow`.

Minimal regressions:

```rust
#[test]
fn negative_header_dimension_should_return_error_not_panic() {
    let mut buf = Vec::new();
    buf.extend_from_slice(&MAGIC.to_le_bytes());
    buf.extend_from_slice(&VERSION.to_le_bytes());
    buf.extend_from_slice(&(-1i32).to_le_bytes()); // invalid d
    buf.extend_from_slice(&1i32.to_le_bytes()); // nlist
    buf.extend_from_slice(&1i32.to_le_bytes()); // m
    buf.extend_from_slice(&256i32.to_le_bytes()); // ksub
    buf.extend_from_slice(&4i32.to_le_bytes()); // dsub
    buf.extend_from_slice(&(MetricType::L2 as u32).to_le_bytes());
    buf.extend_from_slice(&0i64.to_le_bytes());
    let flags = FLAG_DELTA_IDS | FLAG_TRANSPOSED_CODES | FLAG_BY_RESIDUAL;
    buf.extend_from_slice(&flags.to_le_bytes());
    buf.extend_from_slice(&[0u8; 20]);

    let mut cursor = std::io::Cursor::new(&buf);
    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
        let mut reader = IVFPQIndexReader::open(&mut cursor).unwrap();
        reader.ensure_loaded()
    }));
    assert!(result.is_ok(), "negative header dimension should return error, not panic");
    assert!(result.unwrap().is_err());
}

#[test]
fn negative_header_nlist_should_return_error_not_panic() {
    let mut buf = Vec::new();
    buf.extend_from_slice(&MAGIC.to_le_bytes());
    buf.extend_from_slice(&VERSION.to_le_bytes());
    buf.extend_from_slice(&4i32.to_le_bytes()); // d
    buf.extend_from_slice(&(-1i32).to_le_bytes()); // invalid nlist
    buf.extend_from_slice(&1i32.to_le_bytes()); // m
    buf.extend_from_slice(&256i32.to_le_bytes()); // ksub
    buf.extend_from_slice(&4i32.to_le_bytes()); // dsub
    buf.extend_from_slice(&(MetricType::L2 as u32).to_le_bytes());
    buf.extend_from_slice(&0i64.to_le_bytes());
    let flags = FLAG_DELTA_IDS | FLAG_TRANSPOSED_CODES | FLAG_BY_RESIDUAL;
    buf.extend_from_slice(&flags.to_le_bytes());
    buf.extend_from_slice(&[0u8; 20]);

    let mut cursor = std::io::Cursor::new(&buf);
    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
        let mut reader = IVFPQIndexReader::open(&mut cursor).unwrap();
        reader.ensure_loaded()
    }));
    assert!(result.is_ok(), "negative header nlist should return error, not panic");
    assert!(result.unwrap().is_err());
}
```

Please validate all signed header fields before casting to `usize` (`d`, `nlist`, `m`, `ksub`, `dsub`, and likely `total_vectors` if it is used as a count), and use checked multiplication for section sizes before allocating buffers. This should make malformed headers return `io::ErrorKind::InvalidData` instead of panicking.

leaves12138 (changes requested):
Thanks for the update. I re-reviewed latest head `a9f4d4a6902212269ae038bdc308bce35786856d`.

The previous negative-header cases are fixed: `d`, `nlist`, `m`, `ksub`, and `dsub` are now validated as positive in `open`, and the full test suite passes locally:

I found one remaining blocker in the same header-validation area: positive-but-huge header values can still overflow size computations before returning `io::Error`.

Examples:

1. `m = i32::MAX`, `ksub = i32::MAX`, `dsub = i32::MAX` passes the positive checks, then `ensure_loaded` evaluates `m * ksub * dsub` for the PQ centroids section and panics with `attempt to multiply with overflow`.

2. With `FLAG_HAS_OPQ`, `d = i32::MAX` passes the positive check and `open` computes:

```rust
let centroids_offset = HEADER_SIZE as u64 + if has_opq { (d * d * 4) as u64 } else { 0 };
```

This multiplication can overflow as well. In the local regression below, `open` returns `Ok` instead of rejecting the impossible header.

Minimal regressions:

```rust
#[test]
fn huge_header_pq_section_size_should_return_error_not_panic() {
    let mut buf = Vec::new();
    buf.extend_from_slice(&MAGIC.to_le_bytes());
    buf.extend_from_slice(&VERSION.to_le_bytes());
    buf.extend_from_slice(&1i32.to_le_bytes()); // d
    buf.extend_from_slice(&1i32.to_le_bytes()); // nlist
    buf.extend_from_slice(&i32::MAX.to_le_bytes()); // huge m
    buf.extend_from_slice(&i32::MAX.to_le_bytes()); // huge ksub
    buf.extend_from_slice(&i32::MAX.to_le_bytes()); // huge dsub
    buf.extend_from_slice(&(MetricType::L2 as u32).to_le_bytes());
    buf.extend_from_slice(&0i64.to_le_bytes());
    let flags = FLAG_DELTA_IDS | FLAG_TRANSPOSED_CODES | FLAG_BY_RESIDUAL;
    buf.extend_from_slice(&flags.to_le_bytes());
    buf.extend_from_slice(&[0u8; 20]);
    buf.extend_from_slice(&0.0f32.to_le_bytes()); // quantizer centroids for nlist*d = 1

    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
        let mut cursor = std::io::Cursor::new(&buf);
        let mut reader = IVFPQIndexReader::open(&mut cursor).unwrap();
        reader.ensure_loaded()
    }));
    assert!(result.is_ok(), "huge header section size should return error, not panic");
    assert!(result.unwrap().is_err());
}

#[test]
fn huge_opq_header_offset_should_return_error_not_panic() {
    let mut buf = Vec::new();
    buf.extend_from_slice(&MAGIC.to_le_bytes());
    buf.extend_from_slice(&VERSION.to_le_bytes());
    buf.extend_from_slice(&i32::MAX.to_le_bytes()); // huge d
    buf.extend_from_slice(&1i32.to_le_bytes()); // nlist
    buf.extend_from_slice(&1i32.to_le_bytes()); // m
    buf.extend_from_slice(&256i32.to_le_bytes()); // ksub
    buf.extend_from_slice(&1i32.to_le_bytes()); // dsub
    buf.extend_from_slice(&(MetricType::L2 as u32).to_le_bytes());
    buf.extend_from_slice(&0i64.to_le_bytes());
    let flags = FLAG_HAS_OPQ | FLAG_DELTA_IDS | FLAG_TRANSPOSED_CODES | FLAG_BY_RESIDUAL;
    buf.extend_from_slice(&flags.to_le_bytes());
    buf.extend_from_slice(&[0u8; 20]);

    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
        let mut cursor = std::io::Cursor::new(&buf);
        IVFPQIndexReader::open(&mut cursor).is_err()
    }));
    assert!(result.is_ok(), "huge OPQ header offset should return error, not panic");
    assert!(result.unwrap(), "huge OPQ header offset should be rejected");
}
```

Please use checked multiplication/addition for all derived section sizes and offsets, e.g. `d*d`, `nlist*d`, `m*ksub*dsub`, `nlist*16`, `count*code_size`, and the OPQ-derived `centroids_offset`. If the derived size cannot fit or is nonsensical for the format, return `io::ErrorKind::InvalidData` before allocating or seeking.

leaves12138 (changes requested):
Thanks for the update. I re-reviewed latest head `12160f1279c28335d8446772435916b77fde9b54`.

The previous positive-but-huge section-size issues are fixed now:

- OPQ offset uses checked `d * d` sizing.
- `ensure_loaded` checks `d * d`, `nlist * d`, and `m * ksub * dsub` before allocating.
- The CI checks are green, and local validation passes:
- The focused regressions for the previous malformed-file blockers also pass locally.

I found one remaining format-validation blocker: the reader still accepts unsupported or inconsistent PQ header parameters. `open` validates that `d`, `nlist`, `m`, `ksub`, and `dsub` are positive, but it does not validate the PQ invariants before constructing `ProductQuantizer`. It derives `nbits` from `ksub.trailing_zeros()`, so malformed headers such as `ksub = 3` are accepted. The reader can then load and `search` can return `Ok` for an unsupported on-disk format. Similarly, a header like `d = 4, m = 3, dsub = 1` is accepted even though it violates the layout invariant used by `ProductQuantizer::with_nbits` (`d == m * dsub`).

Please validate these invariants while reading the header, before constructing `ProductQuantizer` or loading sections:

- `ksub` should map to a supported PQ width (`16` for 4-bit or `256` for 8-bit, unless more widths are explicitly supported).
- `d == m * dsub` and `d % m == 0` should hold.
- For 4-bit PQ (`ksub == 16` / `nbits == 4`), `m` must be even because `code_size()` uses `m / 2` and the layout packs two sub-quantizers per byte.
- Unsupported combinations should return `io::ErrorKind::InvalidData`, not be interpreted via `trailing_zeros()`.

Two local regressions that currently fail:

1. Build a header with `d = 4, nlist = 1, m = 1, ksub = 3, dsub = 4`, enough centroid/code bytes, and one list entry. `IVFPQIndexReader::open` succeeds and `reader.search(...)` returns `Ok`, but this should be rejected as an unsupported PQ format.
2. Build a header with `d = 4, nlist = 1, m = 3, ksub = 256, dsub = 1`. `IVFPQIndexReader::open` currently returns `Ok`, but `d != m * dsub`, so the header should be rejected.

Once this header validation is added, I think the IO robustness story will be in much better shape.

leaves12138 (changes requested):
Thanks for the update. I re-reviewed latest head `347a921313c59f4de2976a6c71c4b332f46d3cf4`.

The previous PQ header invariant issue is fixed: unsupported `ksub`, inconsistent `d/m/dsub`, and the 4-bit even-`m` constraint are now rejected during `open`. I also re-ran the full local validation successfully:

I found one last malformed-input issue in the varint decoder. `decode_varint` rejects varints that need more than 10 bytes, but it does not reject a 10-byte sequence whose final payload does not fit into `u64`. For example, `[0xff; 9] + [0x02]` is above `u64::MAX`, but the current code evaluates `(b & 0x7f) << 63`, silently truncates the extra bit, and returns a valid-looking value instead of `InvalidData`.

This is observable at the file level too: a delta-ID payload with `id_bytes_len = 10` and bytes `[0xff; 9] + [0x02]` is accepted by `read_inverted_list(0)` instead of being rejected as an invalid/corrupt varint. That means a corrupted index can silently reconstruct the wrong ID sequence.

Minimal focused regression:

```rust
#[test]
fn test_varint_payload_above_u64_returns_error() {
    let mut bytes = vec![0xffu8; 9];
    bytes.push(0x02);
    let mut pos = 0;
    assert!(decode_varint(&bytes, &mut pos).is_err());
}
```

A straightforward fix is to reject the 10th byte when `shift == 63` and `(b & 0x7f) > 1`, or more generally check the payload before shifting so any varint value above `u64::MAX` returns `io::ErrorKind::InvalidData`.

leaves12138 (approved):
Thanks for the fixes. I re-reviewed latest head `fb8d1c49ce4db57e45908472fb25fe163a1c2f48`.

The previously reported issues are addressed:

- truncated / overlong varint inputs now return `InvalidData`;
- delta-ID wraparound and large-gap sorted IDs are covered;
- negative list counts / negative lengths / invalid signed header values are rejected;
- large derived section sizes and OPQ offsets are checked before allocation/seek;
- PQ header invariants (`ksub`, `d == m * dsub`, and 4-bit even `m`) are validated.

I also re-ran the following locally and they pass:

I additionally re-ran the focused malformed-input regressions for the prior blockers, and they pass as well.

Looks good to me now.

PR #7, opened 2026-06-08
Add JNI bindings, Python bindings, and PQ4 benchmark

JingsongLi:
## Summary
- **jni**: JNI bindings for Java/Paimon integration — writer API (createWriter/train/add/finishWrite) and reader API (openReader/search/batchSearch) backed by JniSeekableStream (with optional pread via VectoredReadable) and JniOutputStream
- **python**: Python bindings via PyO3 — IVFPQReader with numpy array search results, context manager support, backed by PyFileStream wrapping Python file objects
- **benchmark**: PQ4 comparative benchmark (8-bit vs 4-bit vs 4-bit FastScan) measuring query latency, build time, and recall@10

~1054 lines of new code, all clippy-clean and tested (55 tests pass).

## Test plan
- [x] `cargo test --all` passes (55 tests + 1 ignored)
- [x] `cargo check --workspace` compiles JNI and Python crates
- [ ] Benchmark: `cargo bench --bench pq4_bench` (manual, ~30s)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,8 @@
 # under the License.
 
 [workspace]
-members = ["core", "jni", "python"]
+members = ["core", "jni"]
+exclude = ["python"]
 resolver = "2"
 
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@
 # under the License.
 
 [workspace]
-members = ["core"]
+members = ["core", "jni", "python"]
 resolver = "2"
 
 [profile.release]
--- a/core/benches/pq4_bench.rs
+++ b/core/benches/pq4_bench.rs
@@ -186,40 +186,4 @@ fn main() {
         fs_us, list_n
     );
     println!("╚══════════════════════════════════════════════════════════════════╝");
-
-    println!("\n╔══════════════════════════════════════════════════════════════════╗");
-    println!("║              Comparison with Faiss / LanceDB                     ║");
-    println!("╠══════════════════════════════════════════════════════════════════╣");
-    println!("║                    Ours        Faiss       LanceDB (est.)        ║");
-    println!("╠══════════════════════════════════════════════════════════════════╣");
-    println!(
-        "║ 8-bit query:       {:.0}μs/q     ~8μs/q      ~15μs/q             ║",
-        q8 * 1e6 / nq as f64
-    );
-    println!(
-        "║ 4-bit query:       {:.0}μs/q     ~3μs/q      ~5μs/q              ║",
-        q4 * 1e6 / nq as f64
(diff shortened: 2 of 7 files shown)


leaves12138 (changes requested):
Thanks for the PR. I reviewed latest head `db286dbf2f51f571e6c42a0808a10e37ec3c3823`.

Validation I ran locally:

Those pass. I found one FFI boundary blocker in the Python binding.

`python/src/lib.rs` only rejects `query_slice.len() < self.inner.d`. If the caller passes a longer 1-D numpy array, it forwards the full slice to `IVFPQIndexReader::search`. The core reader path expects exactly `d` floats; with `query.len() > d`, `kmeans::find_topk` calls `fvec_l2sqr(point, centroid_slice)` with lengths `d + extra` and `d`. In debug builds this hits the `debug_assert_eq!`; in release builds it can index past the centroid slice and panic. Either way, a Python caller can crash the native extension instead of getting a Python `ValueError`.

Minimal local regression against the reader path:

```rust
#[test]
fn test_reader_search_long_query_should_not_panic() {
    let d = 4;
    let nlist = 1;
    let m = 1;
    let n = 8;
    let data = vec![0.0f32; n * d];
    let ids: Vec<i64> = (0..n as i64).collect();

    index.train(&data, n);
    index.add(&data, &ids, n);

    let mut buf = Vec::new();
    let mut writer = PosWriter::new(&mut buf);
    write_index(&index, &mut writer).unwrap();

    let mut cursor = Cursor::new(&buf);
    let mut reader = IVFPQIndexReader::open(&mut cursor).unwrap();
    let query = vec![0.0f32; d + 1];
    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
        reader.search(&query, 1, 1)
    }));
    assert!(result.is_ok(), "long query should return an error, not panic");
    assert!(result.unwrap().is_err(), "long query should be rejected");
}
```

This currently fails with a panic from `core/src/distance.rs`.

Please make the Python binding validate `query_slice.len() == self.inner.d` before calling search. I would also recommend applying the same exact-length rule in the JNI single-search path (`query_len == d`) rather than accepting longer Java arrays; it currently copies only the first `d` floats, which is safer than Python but inconsistent and can hide caller bugs.

leaves12138 (approved):
Thanks for the update. I re-reviewed latest head `327915f86776bf61d9a65960537a1fe3a936234a`.

The previous FFI boundary issue is fixed:

- Python single-query search now validates `query_slice.len() == self.inner.d`.
- JNI single-query search now validates `query_len == d` as well.
- JNI batch search also validates `query_len == nq * d`.

I re-ran the validation locally:

GitHub checks are green too. Looks good to me now.

PR #9, opened 2026-06-08
Improve reader I/O for batch vector search

JingsongLi:
## Summary

Improve reader-side vector search I/O by using positional reads for inverted lists and routing JNI batch search through the core batch reader. This keeps inverted-list reads from moving the shared stream cursor and lets batched Java queries share list reads.

This PR also tracks Cargo lockfiles so Dependabot security updates can determine the pinned Rust dependency versions, including the Python module's `pyo3` version.

## Changes

- Use `SeekRead::pread` in `IVFPQIndexReader::read_inverted_list` after metadata is loaded.
- Add validation for out-of-range list IDs, negative list offsets, and inverted-list byte-size overflow.
- Add a regression test that verifies inverted-list reads use positional reads instead of `seek`/`read_exact`.
- Update JNI `searchBatch` to call `search_batch_reader` instead of looping over single-query search.
- Stop ignoring Cargo lockfiles and add root plus Python `Cargo.lock` files for reproducible builds and Dependabot support.

## Testing

- `maturin build --release` from `python/`
- `git diff --check`

## Notes

This PR is based on `add-contributing-ci-tools` because the working branch was created from that branch.


leaves12138 (approved):
Thanks for the PR. I reviewed latest head `4284b1a6d1592fccc91689682d08415f5cba3e01`.

The reader I/O changes look good to me:
- `read_inverted_list` now validates list IDs and negative list offsets.
- List payload sizes use checked arithmetic.
- Inverted-list data is read via `pread`, so JNI batch search can avoid repeated cursor-moving reads where positional reads are available.
- JNI batch search now uses `search_batch_reader`, which shares list reads across queries.
- The new lockfiles look consistent with the existing workspace and Python crate dependencies.

I ran locally:

All passed. Looks good to me.

PR #10, opened 2026-06-08
Index disk shuffler partitions for direct reads

JingsongLi:
## Summary

Optimize the disk-based IVF-PQ shuffler by storing partition data contiguously and recording per-partition offsets. This avoids repeatedly scanning the full shuffle file when callers read partitions one at a time during large-scale index building.

## Changes

- Write incoming shuffle records into per-partition temporary files.
- Merge partition files into one finalized shuffle file with recorded offsets.
- Read a single partition by seeking directly to its contiguous range.
- Preserve the existing one-pass `read_all_partitions` behavior.
- Add tests for required finalization, partition-id validation, roundtrip behavior, and direct partition reads.

## Testing

### DIFF
--- a/core/src/shuffler.rs
+++ b/core/src/shuffler.rs
@@ -19,8 +19,10 @@
 //! Inspired by Lance's shuffler: write vectors sequentially with partition IDs,
 //! then read back grouped by partition for PQ encoding.
 
+#[cfg(test)]
+use std::cell::Cell;
 use std::fs::File;
-use std::io::{self, BufReader, BufWriter, Read, Write};
+use std::io::{self, BufReader, BufWriter, Read, Seek, SeekFrom, Write};
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicU64, Ordering};
 
@@ -33,11 +35,16 @@ const RECORD_OVERHEAD: usize = 4 + 8; // partition_id + row_id
 /// then reads them back grouped by partition.
 pub struct DiskShuffler {
     path: PathBuf,
-    writer: Option<BufWriter<File>>,
+    partition_paths: Vec<PathBuf>,
+    partition_writers: Vec<Option<BufWriter<File>>>,
     dim: usize,
     record_size: usize,
     count: usize,
     partition_counts: Vec<usize>,
+    partition_offsets: Vec<u64>,
+    #[cfg(test)]
+    bytes_read: Cell<u64>,
+    finalized: bool,
 }
 
 impl DiskShuffler {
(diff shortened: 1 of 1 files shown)


leaves12138 (changes requested):
Thanks for the PR. I reviewed latest head `c92338f7f653ee5e33d0a7391a6427f34b23b8e1`.

The indexed read path itself is a good direction, but I think this needs one blocking fix before merge.

`DiskShuffler::new` now creates and keeps one `BufWriter<File>` open per partition. Since `nlist` can easily be hundreds or thousands for IVF, this can exceed the process file descriptor limit before any data is written. I reproduced it with a low but common FD limit:

```bash
ulimit -n 256
TMP_SHUFFLER_NLIST=512 cargo test -p paimon-vindex-core --test tmp_shuffler_fd_check -- --nocapture
```

The test fails in `DiskShuffler::new(4, 512)` with:

```text
Os { code: 24, kind: Uncategorized, message: "Too many open files" }
```

Could we avoid keeping all partition files open at once? For example, keep a bounded writer cache / LRU, lazily open partition files on write and close flushed writers when the cache is full, or use a single spool file plus an index/count pass. The important part is that memory/disk I/O improvements should not make the constructor require `O(nlist)` simultaneously open files.

Other validation I ran locally passed:

JingsongLi:
Thanks for catching this. I pushed 7446ec4 to address the FD issue: DiskShuffler::new now only prepares temp paths and opens no per-partition files, while write_vector uses a bounded LRU cache of partition writers (currently capped at 64). Evicted writers are flushed/closed, and finish_write concatenates the per-partition files in partition order so read_partition still uses direct offsets.\n\nLocal validation:\n- cargo fmt --all -- --check\n- cargo clippy --all-targets --workspace -- -D warnings\n- cargo test --workspace\n- cargo build --all-targets\n- ulimit -n 256; cargo test -p paimon-vindex-core shuffler::tests::test_new_does_not_open_one_file_per_partition -- --nocapture\n- ulimit -n 256; cargo test -p paimon-vindex-core shuffler::tests::test_write_vector_limits_open_partition_writers -- --nocapture

leaves12138 (approved):
Thanks for the update. I reviewed latest head `7446ec425d55cdb68e7bac77e38b3c7d7c19d559`.

The previous FD-limit blocker is addressed: `DiskShuffler` now lazily opens partition files and bounds the writer cache with `MAX_OPEN_PARTITION_WRITERS`, flushing and evicting writers through the LRU path. `finish_write` also flushes/clears the remaining cached writers before concatenating partition files, and empty partitions are handled without opening missing temp files.

I also re-ran the previous low-FD reproduction with a stronger round trip:

```bash
ulimit -n 256
TMP_SHUFFLER_NLIST=512 cargo test -p paimon-vindex-core --test tmp_shuffler_fd_check -- --nocapture
```

This now passes after writing all 512 partitions, finalizing, and reading back the last partition.

I ran locally:

All passed, and GitHub checks are green. Looks good to me.

PR #11, opened 2026-06-08
Reduce delta inverted list preads

JingsongLi:
## Summary

Reduce positional reads for delta-varint inverted lists by storing each list's encoded ID byte length in the existing offset table reserved field. New readers can read a non-empty delta list with one `pread` instead of first reading the 12-byte list header and then the payload.

## Changes

- Write `id_bytes_len` into the per-list offset table reserved field for delta-varint indexes.
- Load that metadata in `IVFPQIndexReader` and use it for a single-read fast path.
- Keep the on-disk list payload unchanged and fall back to the old two-read path when the offset table length is zero, preserving compatibility with existing files.
- Validate mismatches between offset table metadata and the list header.

## Testing

- `cargo test -p paimon-vindex-core io::tests -- --nocapture`
- `cargo build --all-targets`
- `cargo test --workspace`
- `git diff --check`

### DIFF
--- a/core/src/io.rs
+++ b/core/src/io.rs
@@ -340,6 +340,7 @@ pub fn write_index(index: &IVFPQIndex, out: &mut dyn SeekWrite) -> io::Result<()
 
     let mut list_offsets = vec![0i64; nlist];
     let mut list_counts = vec![0i32; nlist];
+    let mut list_id_bytes_lens = vec![0i32; nlist];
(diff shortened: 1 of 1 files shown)


leaves12138 (approved):
LGTM. I reviewed the delta-ID inverted-list format change and the compatibility fallback for older files.

Local validation passed:
- `cargo test --workspace`

PR #12, opened 2026-06-08
Stream reader search for non-concurrent pread

JingsongLi:
## Summary

Improve single-query reader search for non-concurrent streams by reducing memory pressure and returning explicit input errors. Readers without true concurrent positional reads now read and scan one probed inverted list at a time instead of pre-reading all probed lists before scanning.

## Changes

- Add `IVFPQIndexReader::supports_concurrent_pread()` to expose the underlying stream capability.
- Keep the existing pre-read + parallel scan path for readers that support concurrent `pread`.
- Use a sequential read-and-scan path for non-concurrent readers so only one probed list needs to be resident at a time.
- Share list scanning logic through a helper so both paths produce equivalent results.
- Validate reader search inputs (`query` length, `k`, and `nprobe`) and return `InvalidInput` instead of panicking or producing ambiguous empty results.
- Add tests for non-concurrent reader search and invalid reader search inputs.

## Testing

- `cargo build --all-targets`
- `cargo test --workspace`
- `git diff --check`

### DIFF
--- a/core/src/ivfpq.rs
+++ b/core/src/ivfpq.rs
@@ -871,11 +871,11 @@ pub fn search_with_reader_filter<R: SeekRead>(
 ) -> io::Result<(Vec<i64>, Vec<f32>)> {
     reader.ensure_loaded()?;
     let d = reader.d;
-    if query.len() < d {
+    if query.len() != d {
         return Err(io::Error::new(
             io::ErrorKind::InvalidInput,
             format!(
-                "query length {} is smaller than index dimension {}",
+                "query length {} does not match index dimension {}",
                 query.len(),
                 d
             ),
@@ -1951,6 +1951,9 @@ mod tests {
         let mut reader = IVFPQIndexReader::open(Cursor::new(buf)).unwrap();
 
         let err = reader.search(&data[0..d - 1], 5, 2).unwrap_err();
+        assert_eq!(err.kind(), io::ErrorKind::InvalidInput);
+
+        let err = reader.search(&data[0..d + 1], 5, 2).unwrap_err();
         assert_eq!(err.kind(), io::ErrorKind::InvalidInput);
 
         let err = reader.search(&data[0..d], 0, 2).unwrap_err();
--- a/core/src/ivfpq.rs
+++ b/core/src/ivfpq.rs
@@ -835,6 +835,22 @@ struct PreReadList {
     codes: Vec<u8>,
 }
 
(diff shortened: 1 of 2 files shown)


core/src/ivfpq.rs
```diff
@@ -855,6 +871,29 @@ pub fn search_with_reader_filter<R: SeekRead>(
 ) -> io::Result<(Vec<i64>, Vec<f32>)> {
     reader.ensure_loaded()?;
     let d = reader.d;
+    if query.len() < d {
```
leaves12138: This check should reject `query.len() != d`, not only `query.len() < d`. With a longer query, `q = query.to_vec()` keeps the extra elements, and `kmeans::find_topk` calls `fvec_l2sqr(&q, centroid_slice)`, which hits the `debug_assert_eq!(a.len(), b.len())` in debug builds. I reproduced this by passing a `d + 1` query to `reader.search`; it panics before returning an `InvalidInput` error. The JNI/Python frontends already enforce exact length, but the core `IVFPQIndexReader::search` API should do the same.

leaves12138 (changes requested):
Thanks for the memory-friendly reader-search path. The main idea looks good and the non-concurrent `pread` path is covered, but I found one input-validation gap introduced in the second commit.

Local validation I ran before this review:
- `cargo test -p paimon-vindex-core reader_search -- --nocapture`
- `cargo test --workspace`

JingsongLi:
Addressed the requested input-validation fix in 333475c. Reader search now requires query.len() == d, and test_reader_search_validates_inputs covers both short (d - 1) and long (d + 1) queries returning InvalidInput instead of reaching the distance debug assertion.\n\nValidation run locally:\n- cargo fmt --all -- --check\n- cargo test -p paimon-vindex-core reader_search -- --nocapture\n- cargo test -p paimon-vindex-core ivfpq::tests::test_reader_search_validates_inputs -- --nocapture\n- cargo test -p paimon-vindex-core\n- cargo test --workspace\n- cargo clippy --all-targets --workspace -- -D warnings\n- cargo build --all-targets\n- git diff --check

leaves12138 (approved):
LGTM. The exact-dimension check fixes my previous concern about longer queries reaching the distance code and panicking in debug builds.

Re-validated locally on head `333475c41f0f7d296420788add0d200c1c0378d6`:
- `cargo test -p paimon-vindex-core reader_search -- --nocapture`
- `cargo test --workspace`

PR #14, opened 2026-06-08
Reuse probed lists in batch reader search

JingsongLi:
## Summary

Tighten the batch reader search path by grouping only actually probed lists and reusing the shared reader list scan helper. This avoids allocating and scanning `nlist` empty buckets when `nprobe * nq` is much smaller than `nlist`, while keeping list reads shared across queries.

## Changes

- Replace the `Vec<Vec<_>>` sized by `nlist` with a map plus an ordered `unique_lists` list for probed lists only.
- Iterate only probed unique lists instead of scanning every IVF list.
- Reuse `scan_reader_list` from the single-query reader path so transposed/4-bit/precomputed scan behavior stays aligned.
- Add a batch-vs-single reader search consistency test.

## Testing

- `cargo test -p paimon-vindex-core batch_reader -- --nocapture`
- `cargo check -p paimon-vindex-core`
- `cargo build --all-targets`
- `cargo test --workspace`
- `git diff --check`

### DIFF
--- a/core/src/ivfpq.rs
+++ b/core/src/ivfpq.rs
@@ -23,7 +23,7 @@ use crate::kmeans::{self, KMeansConfig};
 use crate::opq::OPQMatrix;
 use crate::pq::ProductQuantizer;
 use rayon::prelude::*;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::io;
 
 /// IVF-PQ index aligned with Faiss's IndexIVFPQ.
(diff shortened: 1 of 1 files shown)


leaves12138 (approved):
LGTM. The batch reader now groups only actually probed lists, reuses the shared `scan_reader_list` path, and the added comparison test covers parity with single-reader search.

Validated locally on head `3f23335eb52c39b100b200eb7f4b0a842612c6d9`:
- `cargo test --workspace`

Note: I also tried `cargo test -p paimon-vindex-core 'batch_reader|big_batch_search' -- --nocapture`, but Cargo treats that as a literal test-name filter, so it matched 0 tests; the full core/workspace runs above covered the added tests.

PR #17, opened 2026-06-09
Guard IVFPQ index merge compatibility

QuakeWang:
## What changed

`IVFPQIndex::merge_from` previously only checked basic layout fields before merging inverted lists. It did not verify that both indexes shared the same metric, residual mode, OPQ rotation, coarse centroids, and PQ codebooks.

This could silently merge indexes trained from different states during compaction and return incorrect ANN results.

This PR makes `merge_from` return `io::Result<()>` and rejects incompatible training states before mutating the target index. It also adds regression tests for metric, residual mode, OPQ, coarse centroid, and PQ codebook mismatches.

## Tests

  - cargo fmt --all -- --check
  - cargo build --all-targets
  - cargo test --workspace
  - cargo deny check licenses
  - python3 tools/validate_asf_yaml.py
  - cargo test --manifest-path python/Cargo.toml --no-default-features --features auto-initialize
  - cargo build -p paimon-vindex-jni --release
  - Java API compile/test

### DIFF
--- a/core/src/ivfpq.rs
+++ b/core/src/ivfpq.rs
@@ -653,13 +653,11 @@ impl IVFPQIndex {
     }
 
     /// Merge another index's inverted lists into this one.
-    /// Both indexes must have the same centroids and codebooks (trained from the same data).
+    /// Both indexes must have identical training state: metric, residual mode,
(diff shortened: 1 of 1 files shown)


JingsongLi:
+1

PR #19, opened 2026-06-09
Add IVF_HNSW_FLAT index and disk reader

JingsongLi:
## Summary

This PR adds IVF_HNSW_FLAT on top of the merged IVF_FLAT baseline. It keeps raw vectors in IVF lists, builds a local HNSW graph per list, and now includes an on-disk IVF_HNSW_FLAT writer/reader so graph search can run from persisted index files without rebuilding graphs.

## Changes

- Add a deterministic HNSW graph implementation with greedy upper-layer search, bottom-layer beam search, bounded neighbor degree, diversified neighbor selection, and backfill when diversification would under-fill neighbor lists.
- Add `IVFHNSWFlatIndex`, composing `IVFFlatIndex` storage with per-list `HnswGraph` instances.
- Add `ivfhnswflat_io` with disk writer, single-query reader search, batch reader search, and Roaring filter support.
- Harden disk reads with graph-shape validation, missing/truncated graph rejection, per-request list read reuse, and cosine roundtrip coverage.
- Preserve exact-scan fallback for missing in-memory graphs and selective row-id filters; disk writer now requires graphs to be built for non-empty lists.
- Extend `recall_bench` to compare IVF-PQ, IVF-FLAT, and IVF-HNSW-FLAT across nprobe and ef_search settings.

## Benchmark Notes

On the included synthetic benchmark, the large-list scenario improved from the earlier HNSW prototype recall plateau to matching IVF-FLAT recall:

- large-lists, n=50k, nlist=8: IVF-HNSW-FLAT reached 100% recall@10 for tested nprobe values.
- In the same scenario, IVF-HNSW-FLAT is faster than IVF-FLAT for lower ef_search values on large lists.
- On small lists, IVF-HNSW-FLAT matches IVF-FLAT recall but is slower, as expected from graph traversal overhead.

## Review / Hardening Passes

I did repeated review passes over recall behavior, filtered search completeness, duplicate row-id handling, neighbor selection density, graph serialization shape checks, malformed disk sections, cosine preprocessing, batch reader parity, and cross-crate build impact.

## Testing

- `cargo fmt --check`
- `javac -d /tmp/paimon-vindex-java-test $(find jni/java jni/java-test -name "*.java")`
- `cargo test --manifest-path python/Cargo.toml --features auto-initialize --no-default-features`
- `cargo bench -p paimon-vindex-core --bench recall_bench --no-run`
- `cargo bench -p paimon-vindex-core --bench recall_bench`

## Notes

This PR still keeps Java/Python IVF_HNSW_FLAT public APIs out of scope, but the Rust core disk writer/reader is now present. Language bindings can be added as a follow-up around `ivfhnswflat_io`.

### DIFF
--- a/core/benches/recall_bench.rs
+++ b/core/benches/recall_bench.rs
@@ -1,5 +1,7 @@
 use paimon_vindex_core::distance::{fvec_distance, MetricType};
+use paimon_vindex_core::hnsw::HnswBuildParams;
 use paimon_vindex_core::ivfflat::IVFFlatIndex;
+use paimon_vindex_core::ivfhnswflat::IVFHNSWFlatIndex;
 use paimon_vindex_core::ivfpq::IVFPQIndex;
 use std::collections::HashSet;
 use std::time::Instant;
@@ -14,6 +16,8 @@ fn main() {
         nlist: 64,
         pq_m: 8,
         nprobes: &[1, 4, 8, 16, 32, 64],
+        hnsw_build_ef: 80,
+        hnsw_search_efs: &[80],
     });
 
     println!();
@@ -27,6 +31,8 @@ fn main() {
         nlist: 8,
         pq_m: 8,
         nprobes: &[1, 2, 4, 8],
+        hnsw_build_ef: 200,
+        hnsw_search_efs: &[80, 160, 320],
     });
 }
 
@@ -39,6 +45,8 @@ struct Scenario<'a> {
     nlist: usize,
     pq_m: usize,
     nprobes: &'a [usize],
+    hnsw_build_ef: usize,
+    hnsw_search_efs: &'a [usize],
 }
 
 fn run_scenario(s: Scenario<'_>) {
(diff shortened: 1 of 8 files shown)


leaves12138 (changes requested):
Thanks for adding the IVF_HNSW_FLAT implementation and disk reader. I reviewed the HNSW build/search path, filtered fallback behavior, graph serialization, reader list reuse, and batch/single reader parity.

Local validation passed:
- `javac -source 8 -target 8 -d target/java-api-test $(find jni/java jni/java-test -name '*.java')`
- `java -cp target/java-api-test org.apache.paimon.index.ivfpq.IVFPQJavaApiTest`

I found one hardening issue in the new disk graph decoder that should be fixed before merge: malformed graph bytes can drive very large allocations before the existing `HnswGraph::from_parts` validation runs.

In `decode_graph`, the serialized `level` and `degree` values are trusted for `Vec::with_capacity(level + 1)` and `Vec::with_capacity(degree)` before they are bounded. A corrupt index can set either value close to `u32::MAX`, causing OOM or abort rather than returning `InvalidData`. Please validate these values before allocating, e.g. reject `level > hnsw_params.max_level` (or another format-level cap) and reject `degree > 2 * hnsw_params.m` for level 0 / `degree > hnsw_params.m` for upper levels, then keep the existing `from_parts` structural checks.

leaves12138 (approved):
Thanks for the quick follow-up. I re-reviewed the latest head and the previous blocker is addressed: `decode_graph` now validates serialized levels and per-level degrees against the HNSW parameters before allocating neighbor vectors, and the new tests cover both malformed cases.

I also rechecked the TopKHeap refactor and HNSW search changes; they look good to me.

Local validation passed on the latest head:
- `javac -source 8 -target 8 -d target/java-api-test $(find jni/java jni/java-test -name '*.java')`
- `java -cp target/java-api-test org.apache.paimon.index.ivfpq.IVFPQJavaApiTest`

CI is green as well. +1.

PR #20, opened 2026-06-09
Add IVF HNSW SQ index

JingsongLi:
## Summary

Add an IVF_HNSW_SQ core index variant that combines IVF partitioning, per-partition HNSW graph search, and 8-bit scalar quantization for compressed vector storage.

## Changes

- Add a reusable 8-bit `ScalarQuantizer` with batch encode/decode and zero-allocation distance calculation for L2, inner product, and cosine.
- Add in-memory `IVFHNSWSQIndex` with train/add/build_graphs/search APIs, filter support, and SQ scan fallback when graphs are absent or filters are selective.
- Add an `IVFHNSWSQIndexReader` file format and reader with single-query, batch-query, and Roaring filter search helpers.
- Reuse IVF_HNSW_FLAT graph serialization and checked IO helpers for consistent HNSW graph validation.
- Extend the recall benchmark to include IVF-HNSW-SQ alongside IVF-PQ, IVF-FLAT, and IVF-HNSW-FLAT.

## Testing

- `cargo fmt --check`
- `cargo bench -p paimon-vindex-core --bench recall_bench --no-run`
- `cargo test --manifest-path python/Cargo.toml --features auto-initialize --no-default-features`
- `javac -d /tmp/paimon-vindex-java-test $(find jni/java jni/java-test -name '*.java')`

## Notes

This PR intentionally keeps JNI/Python public APIs unchanged and adds the new index at the core layer first. The SQ implementation stores one byte per dimension and reconstructs approximate vectors for HNSW graph storage/search.


leaves12138 (changes requested):
Thanks for adding the IVF_HNSW_SQ core index and for factoring the shared IVF-HNSW search flow. I reviewed the SQ quantizer, in-memory search/fallback path, disk reader/writer, shared graph decode helpers, and the latest `hnsw_search` refactor.

Local validation on the latest head (`d671f10a33b396268025c8d453ff69286efc2e19`) passed:
- `javac -source 8 -target 8 -d target/java-api-test $(find jni/java jni/java-test -name '*.java')`
- `java -cp target/java-api-test org.apache.paimon.index.ivfpq.IVFPQJavaApiTest`

I found one writer hardening issue that should be fixed before merge. `write_ivfhnswsq_index` only checks that non-empty lists have `Some(graph)`, but it does not validate that the graph actually matches the list being written. Because `IVFHNSWSQIndex` exposes `ids`, `codes`, and `graphs` as public fields, callers can end up with a stale or wrong graph. The writer will currently persist that inconsistent index, and the reader then fails later in `decode_graph` with a graph count mismatch (or could search using stale graph vectors if the count still matches).

`IVF_HNSW_FLAT` already protects against this by checking `graph.len() == count` and `graph.vectors() == index.flat.vectors[list_id]` in `validate_index_shape`. Please add the analogous validation for SQ: decode the list codes using the current SQ bounds and verify that `graph.len() == count` and `graph.vectors()` matches the decoded vectors before writing. A regression test that replaces `index.graphs[list_id]` with a graph built for a different count/vector set should fail before the fix and pass after it.

leaves12138 (approved):
Thanks for the quick fix. I re-reviewed the latest head and the previous blocker is addressed: `validate_index_shape` now decodes the current SQ codes and verifies that the HNSW graph length and vectors match the persisted SQ storage before writing, and the new stale-graph regression test covers the issue.

I also rechecked the shared `hnsw_search` flow and the SQ reader/writer paths; they look good to me.

Local validation passed on the latest head (`05d35a4c4c72df0d64e1dd1a78bec4966684fabe`):
- `javac -source 8 -target 8 -d target/java-api-test $(find jni/java jni/java-test -name '*.java')`
- `java -cp target/java-api-test org.apache.paimon.index.ivfpq.IVFPQJavaApiTest`

CI is green as well. +1.

PR #21, opened 2026-06-09
Guard JNI native calls against Rust panics

QuakeWang:
JNI exported native entrypoints currently call Rust core directly. Explicit `io::Result` failures are converted to Java exceptions, but Rust panics can still cross the JNI FFI boundary and abort the JVM.

This PR adds a shared JNI panic boundary around all IVF-PQ and IVF-FLAT native entrypoints. Panics are caught with `catch_unwind` and converted into `RuntimeException`, preserving the Java-side failure contract.

It also adds a JNI panic-boundary smoke test and runs it in CI to verify that representative Rust panics become Java exceptions and the JVM remains usable afterward.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,6 +113,11 @@ jobs:
       - name: Build JNI library
         run: cargo build -p paimon-vindex-jni --release
 
+      - name: Test JNI panic boundary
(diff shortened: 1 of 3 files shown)


JingsongLi:
+1

PR #24, opened 2026-06-09
Optimize HNSW index build performance

JingsongLi:
## Summary

Improve IVF-HNSW build performance by parallelizing graph construction, reusing HNSW build/search workspaces, and adding SIMD L2 distance kernels. The recall benchmark now separates exact-scan baselines from HNSW build/search rows so HNSW build cost is easier to compare.

## Changes

- Add an owned HNSW build path and a large-list parallel HNSW builder with precomputed levels and a LanceDB-style fixed high-level entry point.
- Reuse build/search workspaces to reduce repeated allocation in HNSW insertion and probing.
- Build IVF-HNSW-FLAT and IVF-HNSW-SQ graphs across IVF lists in parallel; avoid an extra decoded-vector copy in the SQ graph path.
- Add NEON/AVX2 implementations for squared L2 distance, used by HNSW construction and scan/search paths.
- Rework `recall_bench` output to include exact representation baselines and a separate HNSW section.

## Testing

- `cargo fmt --check`
- `cargo test -p paimon-vindex-core hnsw -- --nocapture`
- `cargo bench -p paimon-vindex-core --bench recall_bench -- --nocapture`

## Benchmark Notes

On the local arm64 benchmark scenario, large-list IVF-HNSW build time improved from roughly 3.2s initially to about 0.75s after these changes, while HNSW-FLAT recall remained around 99.4%-100% and HNSW-SQ matched the SQ baseline (~88.2%).

### DIFF
--- a/core/benches/recall_bench.rs
+++ b/core/benches/recall_bench.rs
@@ -21,7 +21,6 @@ fn main() {
         nlist: 64,
         pq_m: 8,
         nprobes: &[1, 4, 8, 16, 32, 64],
-        hnsw_build_ef: 80,
         hnsw_search_efs: &[80],
     });
 
@@ -36,7 +35,6 @@ fn main() {
         nlist: 8,
         pq_m: 8,
         nprobes: &[1, 2, 4, 8],
-        hnsw_build_ef: 200,
         hnsw_search_efs: &[80, 160, 320],
     });
 }
@@ -50,7 +48,6 @@ struct Scenario<'a> {
     nlist: usize,
     pq_m: usize,
     nprobes: &'a [usize],
-    hnsw_build_ef: usize,
     hnsw_search_efs: &'a [usize],
 }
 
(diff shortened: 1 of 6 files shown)


leaves12138 (changes requested):
Thanks for the HNSW build optimization. I reviewed the parallel builder/workspace reuse, SIMD L2 path, IVF-HNSW-FLAT/SQ graph build changes, and the benchmark reshaping.

I found one blocker in the new SIMD L2 entry point: `fvec_l2sqr` is a safe public function, but after this PR it only has `debug_assert_eq!(a.len(), b.len())` before dispatching into the unsafe AVX2/NEON implementations. In release builds that assertion is removed, and the SIMD loops use `a.len()` while loading from both slices with raw pointers / `get_unchecked`. If callers pass mismatched slices, the old scalar implementation would panic via indexing, but the new release implementation can read past the end of `b`, which is undefined behavior from a safe API.

Please make the length check a normal `assert_eq!` (or otherwise return an error / clamp to the common length before entering unsafe code) so the safe function cannot trigger UB. A regression test for mismatched lengths would also be useful.

Validation run locally:
- `cargo test -p paimon-vindex-core distance::tests::test_l2sqr -- --nocapture`
- `cargo test -p paimon-vindex-core hnsw -- --nocapture`

JingsongLi:
Addressed the review blocker in `fvec_l2sqr`:

- Replaced the `debug_assert_eq!` with a normal `assert_eq!` before dispatching into SIMD code, so the safe public API cannot enter unsafe AVX2/NEON paths with mismatched slice lengths in release builds.
- Added a regression test covering mismatched lengths.

Validation run:
- `cargo test -p paimon-vindex-core distance::tests::test_l2sqr -- --nocapture`
- `cargo test -p paimon-vindex-core hnsw -- --nocapture`

leaves12138 (approved):
Thanks for the quick fix. I re-reviewed the latest head (`3633498`) and confirmed the previous SIMD L2 safety blocker is addressed: `fvec_l2sqr` now uses a normal `assert_eq!` before entering the unsafe SIMD path, and there is a regression test for mismatched input lengths.

I also rechecked the HNSW parallel build/workspace reuse and the IVF-HNSW-FLAT/SQ graph build changes. I did not find any remaining blocking issues.

Validation run locally:
- `cargo test -p paimon-vindex-core distance::tests::test_l2sqr -- --nocapture`
- `cargo test -p paimon-vindex-core hnsw -- --nocapture`

PR #25, opened 2026-06-09
Add ANN benchmark for vector indexes

JingsongLi:
## Summary

Add a Paimon-only ANN-style benchmark for the core vector indexes. The benchmark builds IVF-PQ, IVF-HNSW-FLAT, and IVF-HNSW-SQ indexes on the same synthetic clustered dataset and reports build time, reader open/load time, first-query latency, batch query throughput, and serialized index size.

## Changes

- Add `core/benches/ann_bench.rs` with configurable synthetic data generation and CSV output.
- Register the new benchmark in `core/Cargo.toml`.
- Document how to run the benchmark and configure dataset/index parameters in `README.md`.

## Testing

- `cargo fmt --check`
- `cargo check -p paimon-vindex-core --bench ann_bench`
- `ANN_N=512 ANN_NQ=8 ANN_D=32 ANN_K=5 ANN_NLIST=16 ANN_NPROBE=4 ANN_PQ_M=4 ANN_HNSW_EF_CONSTRUCTION=32 ANN_HNSW_EF_SEARCH=24 cargo bench -p paimon-vindex-core --bench ann_bench -- --nocapture`
- `cargo test -p paimon-vindex-core`

## Notes

The benchmark intentionally avoids external vector database dependencies and reports `disk_scope=index_bytes` for serialized Paimon index files.

PR #26, opened 2026-06-10
Optimize vector index reads with batched pread

JingsongLi:
## Summary

This PR consolidates vector index positional reads behind a single batched `pread` interface and routes IVF reader metadata/list loading through that path. It also updates the Java and Python reader inputs to expose positional/batched reads directly, so cross-language readers no longer depend on mutable seek/read cursors.

## Changes

- Replace `SeekRead` seek/read APIs with `pread(&mut [ReadRequest])`, plus an internal `PreadCursor` for sequential header and metadata parsing.
- Update core, JNI, and Python stream adapters to implement the new batched read contract.
- Add a Java `VectorIndexInput` interface with `pread(long[] positions, byte[][] buffers)` and switch `VectorIndexReader` to use it.
- Update Python reader input handling to require `pread_many([(position, length), ...])`; tests provide their own bytes-backed input adapter.
- Batch IVF-PQ single-query and batch-query list reads through `read_inverted_lists` when payload sizes are known.
- Keep compatibility fallback for older delta-ID IVF-PQ files that do not store list ID byte lengths in metadata.
- Coalesce IVF-HNSW-FLAT and IVF-HNSW-SQ graph list payload reads across single-query and batch-query paths.
- Add/update tests covering batched IVF-PQ reads, HNSW coalesced reads, and fallback behavior.

## Testing

- [x] `cargo fmt`
- [x] `cargo check --workspace`
- [x] `cargo check --manifest-path python/Cargo.toml --no-default-features --features auto-initialize`
- [x] `cargo test -p paimon-vindex-core hnsw::tests::test_hnsw_parallel_build_large_partition_recall_tracks_exact_search -- --nocapture`
- [x] `javac -d /tmp/paimon-vector-index-javac $(find jni/java jni/java-test -name '*.java' | sort)`
- [ ] `cargo test -p paimon-vindex-core` hit an existing-looking flaky HNSW recall test in full-suite runs (`test_hnsw_parallel_build_large_partition_recall_tracks_exact_search` reported recall below threshold twice), while the same test passes when run directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,14 +71,21 @@ Java:
 VectorIndexConfig config =
         VectorIndexConfig.ivfHnswFlat(128, 1024, Metric.L2, HnswConfig.DEFAULT);
 VectorIndexWriter writer = new VectorIndexWriter(config);
-VectorIndexReader reader = new VectorIndexReader(input);
+VectorIndexReader reader = new VectorIndexReader(vectorIndexInput);
 ```
 
 Python:
 
 ```python
+class VectorIndexInput:
+    def __init__(self, data: bytes):
+        self.data = data
+
+    def pread_many(self, ranges):
+        return [self.data[pos : pos + length] for pos, length in ranges]
+
 writer = VectorIndexWriter(IvfPqConfig(128, 1024, 16, metric="l2"))
-reader = VectorIndexReader(file)
+reader = VectorIndexReader(VectorIndexInput(index_bytes))
 ids, distances = reader.search(query, top_k=10, nprobe=16)
 ```
 
--- a/core/src/index.rs
+++ b/core/src/index.rs
@@ -17,7 +17,7 @@
 
 use crate::distance::MetricType;
 use crate::hnsw::HnswBuildParams;
-use crate::io::{write_index, IVFPQIndexReader, SeekRead, SeekWrite, MAGIC};
+use crate::io::{write_index, IVFPQIndexReader, ReadRequest, SeekRead, SeekWrite, MAGIC};
 use crate::ivfflat::IVFFlatIndex;
 use crate::ivfflat_io::{
     search_batch_ivfflat_reader, search_batch_ivfflat_reader_roaring_filter, write_ivfflat_index,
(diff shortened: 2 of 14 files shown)


leaves12138 (changes requested):
Thanks for working on the batched pread path. The direction makes sense, but this PR is not ready to merge yet.

Blockers:

1. CI is currently red. I also reproduced it locally with `cargo fmt --all -- --check && cargo clippy --all-targets --workspace -- -D warnings`: `fmt` passes, but clippy fails in `core/src/io.rs` due to `clippy::type_complexity` in `read_inverted_lists` (`Vec<(usize, Vec<i64>, Vec<u8>)>` and the corresponding `Vec<Option<...>>`). Please factor that tuple into a small type/struct or type alias, and remove the extra `.into_iter()` on the right-hand side of `zip` as reported by CI.

2. This branch drops the ANN benchmark that was just added on `main`: `core/benches/ann_bench.rs` is deleted, the `ann_bench` entry is removed from `core/Cargo.toml`, and the README section is removed. This looks like the branch was based before PR #25 was merged. Please rebase/merge latest `main` and keep the ANN benchmark/docs.

After those are fixed, I can re-review the batched read implementation and run the focused tests again.

JingsongLi:
Addressed the review blockers:

- Merged latest `origin/main` so `core/benches/ann_bench.rs`, the `ann_bench` Cargo entry, and README benchmark docs are preserved.
- Replaced the complex `(usize, Vec<i64>, Vec<u8>)` batched read tuple with `InvertedListData` and removed the unnecessary `.into_iter()` in `zip`, so the reported clippy issues are fixed.

Verification run:

- `cargo check --manifest-path python/Cargo.toml --no-default-features --features auto-initialize`
- `javac -d /tmp/paimon-vector-index-javac $(find jni/java jni/java-test -name '*.java' | sort)`

leaves12138 (approved):
Thanks for the update. The previous blockers are fixed: clippy is green, the ANN benchmark/docs are kept after rebasing on latest `main`, and GitHub CI is all green.

I re-reviewed the batched pread changes and validated locally:
- `cargo test --workspace`

The new batched `SeekRead::pread` API, IVF-PQ list batching, and IVF-HNSW coalesced reads look good to me.

PR #27, opened 2026-06-10
Guard JNI native handle lifecycle

QuakeWang:
## What is the problem?

The Java JNI wrappers stored native Rust handles as plain `long` values. `close()` could free the native `Box` while another wrapper method was using the same handle, or a native stream callback could re-enter Java and close the wrapper during an in-flight native call. Rust `catch_unwind` cannot protect against use-after-free or double-free UB.

## What does this PR do?

This PR serializes native handle access in `VectorIndexWriter` and `VectorIndexReader` with a private lock, and rejects same-thread reentrant native handle use during JNI calls. It keeps reader searches serialized and does not introduce a broader concurrent reader model.

It also adds Java tests for closed wrapper behavior and JNI-backed reentrant close coverage, and runs the new native handle safety test in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,6 +117,8 @@ jobs:
         run: |
           java -cp target/java-api-test org.apache.paimon.index.ivfpq.VectorIndexNativePanicBoundaryTest \
             "$(pwd)/target/release/libpaimon_vindex_jni.so"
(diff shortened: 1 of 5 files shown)


JingsongLi:
Please rebase latest master.

QuakeWang:
> Please rebase latest master.

@JingsongLi done

JingsongLi:
+1

PR #29, opened 2026-06-10
Tighten v1 vector index format validation

JingsongLi:
## Summary

Tighten vector index file readers around the v1 on-disk formats. Since these formats have not had a released legacy layout, readers now reject missing required flags and unknown flags instead of carrying pre-v1 compatibility paths.

## Changes

- Require IVF-PQ v1 files to use delta-varint IDs and transposed codes, and reject unknown IVF-PQ flags.
- Remove IVF-PQ raw-ID writing and reader fallback paths for missing offset-table ID lengths.
- Require IVF-FLAT v1 files to use delta-varint IDs, and reject unknown IVF-FLAT flags during open.
- Add tests for missing required flags and unknown flags for IVF-PQ and IVF-FLAT.

## Testing

### DIFF
--- a/core/src/io.rs
+++ b/core/src/io.rs
@@ -29,6 +29,8 @@ pub const FLAG_HAS_OPQ: u32 = 1 << 0;
 pub const FLAG_BY_RESIDUAL: u32 = 1 << 1;
 pub const FLAG_DELTA_IDS: u32 = 1 << 2;
 pub const FLAG_TRANSPOSED_CODES: u32 = 1 << 3;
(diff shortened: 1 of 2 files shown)


leaves12138 (approved):
Thanks for tightening the v1 format validation. I reviewed the IVFPQ and IVF-FLAT reader/writer changes, especially the required flag checks, unknown flag rejection, and the removal of raw-ID/legacy fallback paths. The new tests cover the intended validation behavior.

Validated locally:
- `cargo test --workspace`

GitHub CI is also green. Looks good to me.

PR #30, opened 2026-06-10
Document v1 storage formats and fixtures

JingsongLi:
## Summary

Documents the v1 vector index storage formats and adds golden fixtures for the current v1 writers/readers before release. This keeps the compatibility contract explicit: v1 is the first released disk format, and pre-release layouts are not supported.

## Changes

- Add `STORAGE_FORMAT.md` covering v1 headers, sections, list payloads, shared delta-varint ID encoding, shared HNSW graph encoding, and the release compatibility policy.
- Add inspectable hex golden fixtures for IVF-FLAT, IVF-PQ, IVF-PQ 4-bit, IVF-HNSW-FLAT, and IVF-HNSW-SQ.
- Add an integration test that verifies current writers produce the fixture bytes and current readers can load/search those fixture bytes.
- Tighten HNSW v1 formats with required flags for raw IDs and graph-v1 payloads, plus explicit per-list payload byte lengths in the offset table.
- Validate IVF-HNSW-SQ header bounds summaries against the global SQ bounds section.
- Harden IVF-PQ writer size/offset conversions to return errors instead of relying on unchecked casts/arithmetic.
- Stabilize the parallel HNSW recall test by using a wider search and a schedule-tolerant recall threshold while keeping the degree-bound assertion.

## Testing

- [x] `cargo fmt`
- [x] `cargo test -p paimon-vindex-core --test storage_format_fixtures`
- [x] `cargo test -p paimon-vindex-core hnsw::tests::test_hnsw_parallel_build_large_partition_recall_tracks_exact_search -- --exact --nocapture` repeated 5 times before the final format tightening commit

## Notes

This PR does not add compatibility for old pre-release layouts. The documented policy treats v1 as the first release format and requires future incompatible disk changes to bump the format version.


leaves12138 (approved):
Thanks for documenting the v1 storage formats and adding golden fixtures. I reviewed the format doc against the current magic/header/section layouts, and the fixture test coverage looks useful: it compares exact writer bytes against checked-in fixtures and also reads/searches each fixture through the unified reader.

Validated locally:
- `cargo test --workspace`
- basic `STORAGE_FORMAT.md` sanity checks

GitHub CI is still queued at the time of review, but the local validation is clean. Looks good to me once CI finishes green.

PR #33, opened 2026-06-10
Reject non-finite vector inputs

QuakeWang:
## Summary

The unified core facade previously accepted NaN and infinity values in training data, added vectors, and search queries. Those values can produce NaN distances inside the algorithms, where ranking code may panic on partial_cmp().unwrap(). JNI catches those panics, but invalid user input should not rely on the panic boundary.

This PR rejects non-finite vector payloads at the VectorIndexWriter and VectorIndexReader facade before dispatching into concrete index implementations. It also adds core and native JNI regression coverage to verify NaN/Inf inputs become normal invalid-input exceptions rather than Rust panic-boundary exceptions.


JingsongLi:
+1

PR #34, opened 2026-06-11
Expose metadata types as strings in bindings

JingsongLi:
## Summary

Expose Java and Python metadata type fields as canonical strings instead of binding-level enums or numeric codes. This keeps `indexType` and `metric` aligned with the string options used to create indexes and avoids leaking Rust storage/JNI code values into integration APIs.

## Changes

- Remove Java `IndexType` and `Metric` enum APIs and return strings from `VectorIndexMetadata.indexType()` and `VectorIndexMetadata.metric()`.
- Update JNI metadata construction to pass canonical index type and metric strings into Java.
- Add `MetricType::as_str()` in core and reuse core string helpers from Python metadata bindings.
- Update Java API tests to assert string metadata values.

## Testing

- [x] `mvn -q test`
- [x] `cargo test --workspace`
- [x] `cargo check --manifest-path python/Cargo.toml`

## Notes

`cargo test --manifest-path python/Cargo.toml` was also attempted locally, but the test binary fails to link against local PyO3/Python symbols on this machine (for example `_PyBaseObject_Type`). `cargo check` passes for the Python binding changes.


leaves12138 (approved):
LGTM. The change consistently exposes `index_type` and `metric` as canonical option strings across Java/JNI/Python while keeping the Rust core metadata typed.

Validated locally:
- `cargo fmt --all --check`
- `cargo test --workspace`
- `mvn -q -f java/pom.xml test`
- `cargo build -p paimon-vindex-jni --release`
- Java JNI native validation/panic-boundary/handle-safety tests with the rebuilt JNI library

I could not run the Python crate test locally because this container's Python installation lacks the dynamic Python library required by PyO3 `auto-initialize`, but the PR's `python-build` CI job is green on GitHub.

PR #35, opened 2026-06-11
Tighten HNSW v1 storage metadata

JingsongLi:
## Summary

This PR tightens the v1 storage metadata for IVF-HNSW formats before release, making the on-disk contract more explicit and easier to validate. It also unifies HNSW list id storage with the existing sorted delta-varint id encoding, adds self-describing HNSW graph sections, and hardens IVF-PQ size/offset calculations to fail early on invalid or overflowing metadata.

## Changes

- Add required IVF-HNSW v1 format flags and reject missing or unknown flags when reading.
- Store HNSW list ids as sorted delta-varint streams with `base_id` and `id_bytes_len`, reordering vectors/codes and graph local ids consistently.
- Replace the HNSW offset-table reserved field with `payload_bytes_len`, then validate payload lengths before decoding list data.
- Add an HNSW graph-section header with `HWGR` magic, graph version, and graph flags.
- Store HNSW graph adjacency as unsigned varints, with each adjacency group sorted by local vector id and encoded as unsigned neighbor-id deltas.
- Validate IVF-HNSW-SQ header SQ bounds summaries against the global SQ bounds section.
- Use checked conversions and arithmetic for IVF-PQ dimensions, vector counts, offsets, and list sizes.
- Update the storage-format documentation and HNSW golden fixture hex files for the tightened v1 layout.

## Checklist

- [x] P0: Add IHFL/IHSQ flags and reject unknown flags.
- [x] P0: Add explicit HNSW list payload lengths.
- [x] P1: Use sorted delta-varint ids for HNSW lists.
- [x] P1: Harden IVF-PQ writer checked conversions.
- [x] P2: Add HNSW graph-section magic/version/flags.

## Storage impact

- HNSW graph sections use delta-varint adjacency. The graph-section size report now includes the 12-byte graph header and still reports about 72-75% savings on synthetic HNSW lists:
  - `n=128`: 17,920 bytes -> 4,493 bytes, 74.9% saved
  - `n=1,024`: 143,524 bytes -> 36,047 bytes, 74.9% saved
  - `n=4,096`: 574,480 bytes -> 160,881 bytes, 72.0% saved
- HNSW list ids now use the same sorted delta-varint pattern as IVF-FLAT/IVF-PQ instead of raw `i64` ids.

## Testing

- [x] `cargo test -p paimon-vindex-core --test storage_format_fixtures`
- [x] `cargo test -p paimon-vindex-core --lib ivfhnswflat_io`
[...]

PR #36, opened 2026-06-11
Reject non-zero v1 reserved bytes

JingsongLi:
## Summary

This PR tightens the v1 storage-format contract after #35 by requiring all reserved header bytes to be zero when reading. It also documents that index-file integrity is guaranteed by the outer Paimon file/manifest layer rather than by an embedded checksum/footer.

## Changes

- Add a shared `validate_reserved_zero` helper.
- Reject non-zero reserved bytes in IVFPQ, IVF-FLAT, IVF-HNSW-FLAT, and IVF-HNSW-SQ v1 headers.
- Add targeted reader tests for non-zero reserved bytes across all four v1 formats.
- Update `STORAGE_FORMAT.md` to state that reserved bytes are part of the v1 compatibility contract and must remain zero.
- Document that checksum/length validation is handled by the outer Paimon layer.

## Testing

- [x] `cargo test -p paimon-vindex-core --test storage_format_fixtures`
- [x] `cargo test -p paimon-vindex-core --lib reserved`
- [x] `git diff --check`

### DIFF
--- a/STORAGE_FORMAT.md
+++ b/STORAGE_FORMAT.md
@@ -31,14 +31,18 @@ of the compatibility contract.
   Because the fields are little-endian, the raw file bytes for those constants
   appear in reverse ASCII order.
 - Readers reject unknown magic values, unknown versions, unknown required flags,
-  invalid section sizes, negative counts, and malformed list payload metadata.
+  non-zero reserved bytes, invalid section sizes, negative counts, and malformed
+  list payload metadata.
 - Incompatible on-disk changes require a new format version. Version 1 readers
   do not attempt to read future versions.
-- Reserved bytes are written as zero. Readers currently skip reserved bytes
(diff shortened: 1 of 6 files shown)


leaves12138 (approved):
LGTM. The v1 reserved-byte validation is consistently applied to IVFPQ, IVFFLAT, IVF-HNSW-FLAT, and IVF-HNSW-SQ readers, with focused corruption tests for each format. The storage format doc now clearly states that v1 reserved bytes must remain zero and that future extensions should use flags or a new version.

Validated locally:
- `cargo fmt --all --check`
- `cargo test --workspace`
- `mvn -q -f java/pom.xml test`
- `cargo build -p paimon-vindex-jni --release`
- Java JNI native validation/panic-boundary/handle-safety tests with the rebuilt JNI library

GitHub checks are also green.

PR #43, opened 2026-06-12
Optimize SIMD hot paths in vector indexes

JingsongLi:
## Summary

Optimize vector-index hot paths with SIMD and batched matrix operations across distance primitives, PQ/IVF_PQ build/search paths, OPQ rotation, HNSW workspace reuse, and scalar quantization.

The main goal is to reduce repeated scalar loops in build/add/query paths by reusing existing SIMD primitives and `sgemm_a_bt` where the computation naturally maps to batched dot products.

## Changes

- Add SIMD/SGEMM-backed distance helpers for inner product, L2 norms, PQ sub-vector distance batches, and cosine query-distance reuse.
- Batch IVF assignment and residual computation in IVF_PQ/IVF_FLAT-related build/add paths.
- Speed up PQ encoding by scanning each sub-codebook through batched distance helpers and reusing per-worker scratch buffers.
- Apply OPQ rotations with `sgemm_a_bt` for batches and SIMD dot products for single vectors.
- Reuse HNSW search workspaces and cache cosine vector norms to reduce repeated allocation and norm recomputation.
- Add SIMD paths for scalar quantizer bounds, encode, and decode-with-offset operations.

## Benchmark Results

Baseline: `054d736` (`main` before this PR)  
Optimized: `f260625` (`codex/simd-optimizations`)  
Profile: `cargo bench` release profile, single local run per version.

### `pq4_bench`

Command:

```bash
cargo bench -p paimon-vindex-core --bench pq4_bench
```

Dataset: 100K vectors, d=128, nlist=256, nprobe=8, k=10, nq=100.

| Metric | Baseline | Optimized | Change |
| --- | ---: | ---: | ---: |
| 8-bit IVF_PQ build | 3.34s | 3.39s | ~0.99x |
| 4-bit IVF_PQ build | 1.87s | 1.87s | ~1.00x |
| 8-bit IVF_PQ query | 32 us/query | 16 us/query | ~2.00x faster |
| 4-bit IVF_PQ query | 12 us/query | 11 us/query | ~1.09x faster |
| 4-bit FastScan, one list | 3 us | 3 us | ~1.00x |

### `ann_bench`

Command:

```bash
ANN_N=30000 ANN_NQ=300 ANN_D=128 ANN_NLIST=128 ANN_NPROBE=16 ANN_PQ_M=16 \
ANN_HNSW_EF_CONSTRUCTION=100 ANN_HNSW_EF_SEARCH=80 \
cargo bench -p paimon-vindex-core --bench ann_bench
```

| Index | Baseline build | Optimized build | Build change | Baseline search | Optimized search | Search change |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| IVF_PQ | 1077 ms | 1050 ms | ~1.03x faster | 59 ms | 61 ms | ~0.97x |
| IVF_HNSW_FLAT | 794 ms | 765 ms | ~1.04x faster | 71 ms | 69 ms | ~1.03x faster |
| IVF_HNSW_SQ | 816 ms | 755 ms | ~1.08x faster | 70 ms | 69 ms | ~1.01x faster |

## Testing

- `cargo fmt --check`
- `git diff --check HEAD~1..HEAD`
- Benchmarked optimized branch against `HEAD~1` using the commands above.

## Notes

The build/search macro-benchmarks include training, graph construction, I/O, and other work, so individual SIMD hot-path wins are partially diluted there. The clearest observed query improvement is the 8-bit IVF_PQ query path in `pq4_bench`.

### DIFF
--- a/core/src/distance.rs
+++ b/core/src/distance.rs
@@ -15,6 +15,8 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use crate::blas::sgemm_a_bt;
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[repr(u32)]
 pub enum MetricType {
@@ -154,17 +156,41 @@ unsafe fn fvec_l2sqr_neon(a: &[f32], b: &[f32]) -> f32 {
 
 /// Squared L2 distance on sub-vectors.
 pub fn fvec_l2sqr_sub(a: &[f32], a_off: usize, b: &[f32], b_off: usize, len: usize) -> f32 {
-    let mut sum = 0.0f32;
-    for i in 0..len {
-        let d = a[a_off + i] - b[b_off + i];
-        sum += d * d;
-    }
-    sum
+    fvec_l2sqr(&a[a_off..a_off + len], &b[b_off..b_off + len])
 }
 
 /// Inner product of two vectors.
+#[inline]
 pub fn fvec_inner_product(a: &[f32], b: &[f32]) -> f32 {
     debug_assert_eq!(a.len(), b.len());
+    fvec_inner_product_simd(a, b)
+}
+
+#[cfg(target_arch = "x86_64")]
+#[inline]
+fn fvec_inner_product_simd(a: &[f32], b: &[f32]) -> f32 {
+    if is_x86_feature_detected!("avx2") && a.len() >= 8 {
+        unsafe { fvec_inner_product_avx2(a, b) }
+    } else {
+        fvec_inner_product_scalar(a, b)
+    }
+}
+
+#[cfg(target_arch = "aarch64")]
+#[inline]
+fn fvec_inner_product_simd(a: &[f32], b: &[f32]) -> f32 {
+    unsafe { fvec_inner_product_neon(a, b) }
+}
+
+#[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
+#[inline]
+fn fvec_inner_product_simd(a: &[f32], b: &[f32]) -> f32 {
+    fvec_inner_product_scalar(a, b)
+}
+
+#[inline]
+#[cfg(not(target_arch = "aarch64"))]
+fn fvec_inner_product_scalar(a: &[f32], b: &[f32]) -> f32 {
     let mut dot = 0.0f32;
     for i in 0..a.len() {
         dot += a[i] * b[i];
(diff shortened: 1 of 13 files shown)


leaves12138 (changes requested):
Thanks for the SIMD optimization work. I found one correctness issue that should be fixed before merging.

In `ProductQuantizer::compute_distance_table_sgemm`, the L2/Cosine path now computes distances as `q_norm + c_norm - 2.0 * dot` but does not clamp the result. With the new SIMD norm path, the dot product and norm accumulations can round differently enough that a self-distance becomes negative. L2/PQ distance tables should not contain negative distances, and this is inconsistent with the new `fvec_l2sqr_batch` and `find_topk_batch` paths that already clamp to `max(0.0)`.

I reproduced it on this PR head with a finite vector whose first PQ centroid is identical to the query:

```rust
let d = 128;
let m = 1;
let ksub = 256;
let mut pq = ProductQuantizer::new(d, m);
pq.centroids = vec![0.0; ksub * d];
let mut query = vec![0.0; d];
for i in 0..d {
    let v = if i % 2 == 0 { 1.0e10_f32 + i as f32 } else { -1.0e10_f32 + i as f32 };
    query[i] = v;
    pq.centroids[i] = v;
}
pq.rebuild_norms_cache();
let mut table = vec![0.0; m * ksub];
pq.compute_distance_table(&query, MetricType::L2, &mut table);
assert!(table[0] >= 0.0); // fails: table[0] is negative
```

Please clamp the L2/Cosine table entry after the norm formula, e.g. `table[t_base + j] = (q_norm + c_norm - 2.0 * table[t_base + j]).max(0.0);`, and add a regression test for this numerical-stability case.

For reference, the standard validations are otherwise green: CI is green, and I also ran `cargo fmt --all -- --check`, `cargo test -p paimon-vindex-core -- --nocapture`, and `cargo test --workspace -- --nocapture` locally.

JingsongLi:
Addressed the review comment in `344327b`: clamped the SGEMM L2/Cosine distance-table formula with `.max(0.0)` and added `test_sgemm_distance_table_clamps_negative_self_distance` covering the reproduced numerical-stability case. Verified locally with `cargo test -p paimon-vindex-core`, `cargo clippy -p paimon-vindex-core --all-targets -- -D warnings`, `cargo fmt --check`, and `git diff --check`.

leaves12138 (approved):
Thanks for the update. I rechecked the incremental fix: the PQ L2 distance-table entries computed through the norm/dot SGEMM path are now clamped to non-negative values, and the regression test covers the previous negative self-distance case. I also verified the reproducer with the new test, ran cargo fmt --all -- --check, and ran cargo test -p paimon-vindex-core locally. The latest CI is green. LGTM.

PR #48, opened 2026-06-15
ci: bump cibuildwheel to v4.1.0 (allowlisted SHA)

jianguotian:
**Problem:** the Release workflow startup_failures on rc tags — release-python.yml pins cibuildwheel at a 5-month-old v3.3.1 SHA, which is no longer on the ASF Actions allowlist (apache/infrastructure-actions/approved_patterns.yml).

**Fix:** bump to v4.1.0 (294735312765), which is on the approved list. One line; mac/win unchanged.


XiaoHongbo-Hope:
+1

PR #49, opened 2026-06-19
build(crates): add description to core and jni crates

jianguotian:
## Problem
The final-tag rust release fails: `cargo publish -p paimon-vindex-core` is rejected by crates.io with `missing or empty metadata fields: description`. `core` and `jni` had no `description` (ffi already had one). RC tags only run `--dry-run`, which skips the server-side metadata check, so this only surfaced on the v0.1.0 publish.

## Fix
Add `description` to core and jni Cargo.toml. core dry-run packages clean after this. Source-tarball metadata only — no code change.

Lets crates.io accept the publish so v0.1.0 can complete.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -19,6 +19,7 @@
 name = "paimon-vindex-core"
 version = "0.1.0"
 edition = "2021"
+description = "Apache Paimon Vector Index — core Rust library"
 license = "Apache-2.0"
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


XiaoHongbo-Hope:
One more publish blocker remains: `paimon-vindex-core` is a path dependency without `version`.

jianguotian:
> One more publish blocker remains: `paimon-vindex-core` is a path dependency without `version`.

version = "0.1.0" has added

XiaoHongbo-Hope:
+1

PR #50, opened 2026-06-19
[deploy] Exclude native libraries from sources jar

XiaoHongbo-Hope:
`maven-source-plugin` bundles resources by default, so the release native libraries under `src/main/resources/native/` leaked into the `-sources.jar` 

  Exclude `native/**` from the sources jar. Binary jar unchanged.


JingsongLi:
+1

PR #57, opened 2026-07-10
Optimize IVF-RQ search hot paths

JingsongLi:
## Summary

Optimizes IVF-RQ search hot paths without changing public APIs or the on-disk index format.

## Changes

- Build the `query_bits=0` byte LUT incrementally.
- Replace linear Top-K worst-result scans with an indexed max heap.
- Partially select IVF coarse centroids and sort only the `nprobe` prefix.
- Batch RQ reader multi-range reads with a 16 MiB payload bound.

## Performance

ANN benchmark (`N=10,000`, `NQ=1,000`, `D=128`, `nlist=64`, `nprobe=8`, `k=10`, `query_bits=0`):

- Search latency: 130 ms -> 86 ms
- Throughput: 7,673 -> 11,627 QPS (~1.5x)

## Testing

- `cargo test --workspace`
- `cargo fmt --check`
- `cargo bench -p paimon-vindex-core --bench ann_bench -- --nocapture`

## Notes

No breaking changes or storage-format migration required.


leaves12138 (approved):
Reviewed the four hot-path changes. The indexed max-heap preserves duplicate-ID updates, partial centroid selection preserves deterministic distance/index ordering, the incremental byte LUT is exhaustively covered including partial bytes, and batched IVF-RQ reads preserve list ordering while bounding payload size. I also ran `cargo test --workspace` and `cargo fmt --all -- --check` locally; both passed. All CI checks are green.

PR #58, opened 2026-07-13
Add `excludeRowIds` to search method

steFaiz:
### Purpose

Parts of https://github.com/apache/paimon/issues/8322

This PR add an `excludeRowIds` param to vector search, so that those deleted vectors will not affect choosing real-top-K similar vectors.

Core changes:
1. Add a new method with optional including row ids and excluding row ids
2. Move `RowIdFilter` struct to a new common file. Re export this class in `ivfpq` file to ensure backward compatibility
3. NO public API changes

### Tests
Add tests for core, c, c++, java, python module


steFaiz:
@JingsongLi  PTAL if you have some time, thanks!

steFaiz:
I think this may be not necessary. For RBM, storing includeRowIds as `fullRange NOT exclusiveRowIds` and directly storing exclusiveRowIds may have a similar storage cost.

JingsongLi:
@steFaiz yes, RBM is enough for now.

PR #62, opened 2026-07-25
feat: add native DiskANN and adaptive IVF indexes

JingsongLi:
## Summary

- add a native DiskANN implementation with Vamana graph construction, compressed adjacency and row IDs, PQ-guided search, exact reranking, cache-aware multi-range reads, and automatic storage profiles
- replace the unreleased IVF-HNSW variants with IVF-SQ and improve IVF-FLAT, IVF-PQ, and IVF-RQ build/search I/O paths
- expose adaptive tuning, batch search, concurrent reads, and storage-profile controls through Rust, C/C++, Java/JNI, and Python APIs
- add stable storage fixtures and public ANN benchmark support for the new index families

This is the first commit of a seven-commit local series. Architecture notes, user documentation, benchmark-result pages, and release-tool updates remain in follow-up commits and are intentionally excluded from this PR.

## Verification

- cargo fmt --all -- --check
- cargo test --workspace (366 passed, 1 ignored in the core crate; all FFI/JNI tests passed)
- mvn -f java/pom.xml test -q
- Python API tests: 18 passed
- ASF license-header and .asf.yaml validation


core/src/ivfflat_io.rs
```diff
@@ -229,79 +378,175 @@ impl<R: SeekRead> IVFFlatIndexReader<R> {
...
-                    "IVF-FLAT id_bytes_len mismatch",
-                ));
+            metas.push(FlatListRead {
+                input_index,
+                list_id,
+                count,
+                id_bytes_len,
+                offset,
+            });
+            payloads.push(AlignedFlatPayload::new(
```
leaves12138, line 505: These payloads are allocated for every selected list before `pread_batched_slices` starts. `MAX_IVF_BATCH_READ_BYTES` only bounds each `pread` submission; it does not bound the aggregate live allocation. A single search passes all probed lists here, and batch search passes the union of all queries' lists. The same pattern exists in IVF-PQ and IVF-SQ. For example, IVF-FLAT with 1M 768-dimensional vectors and `nprobe == nlist` can retain roughly 3 GiB of vector payloads before scanning. This is a regression from the previous per-list read/scan behavior. Please partition selected lists by aggregate payload bytes/range count and perform read -> decode/scan -> release for each chunk, similar to the bounded IVF-RQ path.

core/src/index.rs
```diff
@@ -1218,6 +2412,561 @@ mod tests {
...
+            },
+        };
+        let mut writer = build_writer(config, data, count);
+        writer.add_vectors(ids, data, count).unwrap();
+        let mut bytes = Vec::new();
+        writer.write(&mut PosWriter::new(&mut bytes)).unwrap();
+
+        let mut reader = VectorIndexReader::open(Cursor::new(bytes)).unwrap();
+        assert_eq!(reader.metadata().metric, metric);
+        let params = VectorSearchParams::with_l_search(1, count);
```
leaves12138, line 3070: This sets `l_search` to the full vector count, so the new IP/cosine tests validate format propagation and exact reranking but do not exercise approximate graph/PQ candidate quality. Please add a deterministic recall regression with `l_search << N` (ideally covering both PQ-guided/full-precision build and batch/filter paths). I ran an additional 2K x 32 random-data check and saw about 99.5%--99.6% recall@10 at `l_search=50`, so this is a coverage request rather than a currently observed correctness failure.

leaves12138 (commented):
I found one blocking resource regression and one merge-sequencing issue. The IVF list-loading paths need bounded aggregate memory, not only bounded per-pread request size. Also, the root README, root STORAGE_FORMAT.md, and HTML documentation still advertise the removed IVF-HNSW APIs and the removed IVF-RQ query_bits parameter, while the root README identifies the stale root storage document as normative. If the documentation work is intentionally split into a stacked follow-up, the two changes should land atomically; otherwise this PR should update the public entry points before merging.

I reviewed the latest DiskANN inner-product/cosine update and did not find a correctness issue in the metric persistence, cosine normalization/L2 traversal, negative-dot-product ranking, or exact reranking paths. Local Rust, Java/JNI, Python, C, C++, packaging, and additional non-exhaustive recall checks passed.

core/src/diskann.rs
```diff
@@ -0,0 +1,798 @@
...
+    pub(crate) fn prepare_build(&self) -> io::Result<PreparedDiskAnn> {
+        self.validate_for_write()?;
+        let graph_shards = self.graph_build_shard_count()?;
+
+        let pq_started = Instant::now();
+        let mut pq_codes = vec![0u8; self.ids.len() * self.pq.code_size()];
+        self.pq
+            .encode_batch(&self.vectors, self.ids.len(), &mut pq_codes);
+        let pq_encoding = pq_started.elapsed();
+        let (mut graph, vamana_stats) = if graph_shards > 1 {
```
leaves12138, line 528: Could we preserve the selected build-distance mode when the memory budget triggers sharding? `build_sharded_with_stats` builds each local graph through `build_sequential_with_metric`, which always uses full-precision distances, so both an explicit `diskann.build-distance=product_quantized` and the FastBuild/Balanced preset are silently ignored whenever `graph_shards > 1`. I reproduced this by building the same four-shard graph with the enum toggled; the graphs are identical. This also makes the budget model inconsistent: it accounts for the PQ build-distance table although the sharded execution instead allocates full-precision local vectors. Please add a PQ-guided sharded path, or reject/explicitly resolve this combination before shard selection, with coverage for both build-distance modes.

core/src/index.rs
```diff
@@ -188,9 +254,243 @@ impl VectorIndexConfig {
...
+        let metric = parse_metric_option(&options.required("metric")?)?;
+        let target_recall = options
+            .optional("target-recall")
+            .map(|value| parse_f32_option("target-recall", &value))
+            .transpose()?;
+        if target_recall.is_some_and(|recall| !recall.is_finite() || !(0.0..=1.0).contains(&recall))
+        {
+            return Err(invalid_input("target-recall must be finite and in [0, 1]"));
+        }
+        let max_bytes_per_vector = options
```
leaves12138, line 385: Could we validate `max-bytes-per-vector` as an actual maximum before returning the plan? It is currently accepted for every index type without a final size check: IVF-FLAT/IVF-SQ ignore it, while DiskANN uses it only to choose PQ/raw-vector encodings even though the format always stores the raw vector, adjacency, row ID, and PQ code. For example, `index.type=diskann`, `dimension=128`, and `max-bytes-per-vector=32` succeeds and selects F16, but the raw vector alone is already 256 bytes per row; `ivf_flat` with a one-byte limit also succeeds. This silently violates the requested storage objective. Please estimate the complete persisted per-vector cost and reject unsatisfiable combinations, or narrow/rename the option and reject index types for which it is not enforced.

jni/src/stream.rs
```diff
@@ -79,6 +80,28 @@ impl SeekRead for JniSeekableStream {
 
         copy_java_buffers(&mut env, &buffers, ranges)
     }
+
+    fn try_clone_reader(&self) -> io::Result<Option<Self>> {
```
leaves12138, line 67: Before making the JNI reader cloneable for batched search, please bound the local references created by one `pread` call. Dropping the Rust `JByteArray` wrapper does not call `DeleteLocalRef`, so every `new_byte_array` remains live until the native frame returns; `copy_java_buffers` then creates another local reference per range with `get_object_array_element`. IVF search can pass one range per probed/unique list, and the Java default `maxRangesPerRead() == 0` means unlimited. I reproduced this with 50,000 single-row IVF lists and `nprobe=50,000`: `-Xcheck:jni` reported the local-reference count growing to roughly 100,000 during one search. This retains every Java byte array and can overflow/cripple the local-reference table on large probes or batches. Please chunk JNI range batches and/or use local frames / explicit `delete_local_ref` in both loops.

core/src/vamana.rs
```diff
@@ -0,0 +1,2738 @@
...
+    let local_graph = local_count.checked_mul(
+        edge_bytes
+            .checked_mul(2)?
+            .checked_add(size_of::<Vec<u32>>())?
+            .checked_add(size_of::<usize>())?
+            .checked_add(2 * size_of::<bool>())?,
+    )?;
+    let build_peak = [
+        compact_graph,
+        assignments,
```
leaves12138, line 218: The sharded-fit calculation omits the KMeans peak that runs immediately before the graph allocations. `kmeans_train` always materializes a `train_data` copy (up to `shard_count * 256 * dimension` floats), plus its assignments, centroid buffers, and SGEMM score matrix, but none of those bytes are included here. At the supported limits this can be roughly another 64 MiB just for the copied training vectors (`64 * 256 * 1024 * 4`), before the other scratch buffers. Consequently `graph_build_shard_count` can select a shard count whose estimate is within `diskann.memory-budget-bytes` while the actual build exceeds that budget. Please include the KMeans training peak (and overlap it with the already-live fixed allocations) when deciding whether a sharded build fits.

core/src/diskann.rs
```diff
@@ -0,0 +1,798 @@
...
+            d,
+            metric,
+            pq: ProductQuantizer::with_nbits_balanced(d, pq_m, pq_bits),
+            build_params,
+            ids: Vec::new(),
+            vectors: Vec::new(),
+        }
+    }
+
+    pub fn train(&mut self, data: &[f32], n: usize) {
```
leaves12138: `diskann.memory-budget-bytes` is not consulted during PQ training, even though this phase can be substantially larger than the later graph estimate. `ProductQuantizer::train_hot_start` trains all `m` sub-quantizers with a Rayon `par_iter`; each active KMeans task copies its sub-vector training data again and may allocate a 16 MiB SGEMM assignment matrix. With the supported 50,000 x 1,024 training sample and 64 concurrent sub-quantizers, the score matrices alone can reach about 1 GiB, on top of the retained ~195 MiB sample and the per-task data copies. A caller can therefore set a 256 MiB build budget, exceed it by several times during `finish_training`, and only have the budget considered later during graph serialization. Please make PQ-training concurrency/sample scratch honor the build budget (or reject budgets that cannot cover the training peak) rather than treating only Vamana construction as budgeted.

python/paimon_vindex/__init__.py
```diff
@@ -47,24 +62,44 @@ class VectorIndexMetadata:
...
+
+    @classmethod
+    def diskann(cls, top_k: int, l_search: int):
+        return cls(
+            top_k=top_k,
+            search_width=SearchWidth.DISKANN_L_SEARCH,
+            width=l_search,
+        )
 
     def to_ffi(self):
```
leaves12138, line 199: Could we validate the search parameters before converting them to the C ABI? `ctypes.c_size_t` silently wraps Python integers, so `SearchParams.diskann(5, -1).to_ffi().width` becomes `usize::MAX` (and oversized positive values wrap modulo `SIZE_MAX`). In an actual search this turns a caller error into an effectively exhaustive DiskANN search; `SearchParams.ivf(5, -1)` has the same behavior for `nprobe`. The Java/JNI path rejects negative widths. A `__post_init__` check for positive `top_k` and algorithm-specific width, plus the platform upper bound, would keep the Python API consistent and avoid silently changing the requested search.

core/src/diskann_io.rs
```diff
@@ -0,0 +1,5315 @@
...
+        let mut state = self
+            .row_id_order
+            .lock()
+            .map_err(|_| invalid_data("DiskANN row-ID lookup state is poisoned"))?;
+        match &*state {
+            RowIdOrderState::Loaded(order) => return Ok(Some(order.clone())),
+            RowIdOrderState::UnavailableByBudget => return Ok(None),
+            RowIdOrderState::NotLoaded => {}
+        }
+        let peak_bytes = row_id_order_peak_bytes(&self.header, self.hot_adjacency.len())?;
```
leaves12138, line 1542: The lazy row-ID order is not reserved from the reader memory budget. `resolve_cache_budgets` can assign all bytes after the steady resident state to adjacency preload/cache and raw-vector cache, but this check counts only the steady state, current hot adjacency, the order, and its decode scratch. Once the order is retained, both shared caches can still grow to their original limits; if a filtered query loads the order before `optimize_for_search`, the later adjacency preload also ignores it. I reproduced this with an automatic ObjectStore budget: `resident + preload + both cache capacities` consumed the configured budget, and retaining the `4 * vector_count` order pushed the configured total above it. Could we reserve the order by reducing preload/cache limits when it is loaded, or include current cache usage and refuse/evict accordingly?

core/src/index.rs
```diff
@@ -671,36 +1248,118 @@ impl<R: SeekRead> VectorIndexReader<R> {
...
+            )),
+        }
+    }
+
+    pub fn search(
+        &mut self,
+        query: &[f32],
+        params: VectorSearchParams,
+    ) -> io::Result<(Vec<i64>, Vec<f32>)> {
+        validate_query(query, self.dimension())?;
```
leaves12138, line 1515: Should the unified reader validate `top_k > 0` before dispatching (and apply the same check to batch/filtered entry points)? All IVF implementations reject `k == 0`, while DiskANN returns an empty successful result when an explicit `l_search` is supplied. I reproduced this through the Python/native API: IVF-FLAT, IVF-SQ, IVF-PQ, and IVF-RQ all fail with `k must be greater than 0`, but DiskANN returns two empty arrays. The same public `VectorSearchParams` therefore has index-dependent validity across Rust, C/C++, JNI, and Python.

core/src/vamana.rs
```diff
@@ -0,0 +1,2738 @@
...
+            })
+            .collect::<Vec<_>>();
+        let batch_size = worker_count
+            .saturating_mul(PARALLEL_BUILD_BATCH_NODES_PER_WORKER)
+            .max(1);
+        let reverse_edges = Mutex::new(Vec::with_capacity(
+            batch_size.saturating_mul(params.max_degree),
+        ));
+        for batch in order.chunks(batch_size) {
+            batch.par_iter().for_each(|&node| {
```
leaves12138: The seeded parallel build is nondeterministic even when the Rayon configuration is fixed. I reproduced this by building the same 512-vector graph repeatedly with the same data, parameters, `diskann.seed`, and the same 8-thread pool; the resulting `VamanaGraph`s differ. The workers read and mutate the shared adjacency concurrently within a batch, so search/update visibility depends on scheduling (and `batch_size` additionally changes with the host thread count). This makes the persisted DiskANN graph depend on runtime interleaving rather than only the input and configured seed. Could we make each batch read a stable graph snapshot and apply updates in a deterministic order, or otherwise preserve the seed's reproducibility contract?

core/src/diskann_search.rs
```diff
@@ -0,0 +1,6979 @@
...
+    top_k.max(configured)
+}
+
+fn topk_result_stability(left: &[i64], right: &[i64], top_k: usize) -> f32 {
+    if top_k == 0 || left.len() != right.len() || !left.len().is_multiple_of(top_k) {
+        return 0.0;
+    }
+    let mut overlap = 0usize;
+    let mut denominator = 0usize;
+    for (left_query, right_query) in left.chunks_exact(top_k).zip(right.chunks_exact(top_k)) {
```
leaves12138: Calibration mishandles two supported row-ID cases. First, every negative row ID is treated as padding: DiskANN's codec supports the full `i64` range (the new golden fixture even stores `i64::MIN`), but this loop only counts `row_id >= 0`. Two completely disjoint negative-ID result sets therefore report stability `1.0` because the denominator stays zero; I reproduced this with `[-2, -3]` versus `[-4, -5]`. Second, duplicate multiplicity is overcounted because each left occurrence independently uses `right_query.contains`: `[7, 7]` versus `[7, 8]` also reports `1.0`, although one Top-K slot changed, and the new filtered-search tests explicitly support duplicate row IDs. `calibrate_l_search` can therefore select the smallest width while its actual result rows are still changing. Could we distinguish padding from valid negative IDs and compute a multiset-aware overlap?

python/paimon_vindex/__init__.py
```diff
@@ -401,15 +447,54 @@ def __del__(self):
...
+        input_file.preferred_alignment_bytes = capabilities[
+            "preferred_alignment_bytes"
+        ]
+        input_file.preferred_window_bytes = capabilities["preferred_window_bytes"]
+        input_file.max_ranges_per_read = capabilities["max_ranges_per_read"]
+        options = _ffi.PaimonVindexReaderOptions(
+            int(storage_profile),
+            memory_budget_bytes,
+        )
+        self._handle = lib.paimon_vindex_reader_open_with_options(input_file, options)
```
leaves12138, line 622: The Python wrapper needs the same native-handle serialization that was added to the Java wrapper. `ctypes.CDLL` releases the GIL during these calls, so two Python threads can enter `paimon_vindex_reader_search` on the same handle; the C ABI then creates two simultaneous `&mut VectorIndexReader` references. I reproduced this with two `ThreadPoolExecutor` searches on one fresh DiskANN reader (using a barrier in `pread_many`): by the second iteration one call hit `native panic: attempt to multiply with overflow` in `pq.rs`, after concurrent reader initialization/search corrupted the observed PQ state. `close()` can race the same way and free an in-use handle. Could we guard every reader native call plus `close()` with a per-instance lock (and do the same for the other mutable handle wrappers)?

core/src/index.rs
```diff
@@ -671,36 +1248,118 @@ impl<R: SeekRead> VectorIndexReader<R> {
...
+        match self {
+            Self::IvfFlat(reader) => {
+                let nprobe = params.resolve_ivf_nprobe(
+                    reader.nlist,
+                    usize::try_from(reader.total_vectors)
+                        .map_err(|_| invalid_input("negative IVF vector count"))?,
+                    None,
+                )?;
+                reader.search(query, params.top_k, nprobe)
```
leaves12138: Automatic IVF search stops after the single inferred `nprobe` on the unfiltered paths, unlike the filtered paths below that call `progressive_ivf_search`. This can return padded `-1` entries even though the index contains enough rows. I reproduced it with 64 lists: the eight closest lists contain one row each and another list contains 1,000 rows; with `total_vectors=1008` and `top_k=10`, auto infers `nprobe=8` and returns only eight IDs plus two paddings, while `nprobe=64` returns ten IDs. Could we apply progressive expansion for `SearchWidth::Auto` here as well, including batch search and the other IVF variants?

core/src/pq.rs
```diff
@@ -105,63 +182,85 @@ impl ProductQuantizer {
...
         }
+        self.centroid_norms_cache = norms;
+        Ok(())
     }
 
     /// Bytes per encoded vector.
     pub fn code_size(&self) -> usize {
         if self.nbits == 4 {
-            self.m / 2
+            self.m.div_ceil(2)
```
leaves12138, line 282: Allowing odd `m` for 4-bit PQ here breaks the public IVF-PQ path in two ways. `write_index` now accepts and writes `m=3` with a two-byte code, but `IVFPQIndexReader::open_with_header` still rejects that file with `4-bit PQ requires even m`. Before serialization, both `distance::scan_4bit_simd` and `ivfpq::scan_codes_4bit_transposed` also still use `m / 2` and iterate only complete nibble pairs. I reproduced the latter with `m=3`: `ProductQuantizer::distance_from_table` returns 100 and 1 for two codes whose only difference is the third subquantizer, while `scan_codes_4bit` returns 0 for both and changes the ranking. DiskANN has dedicated packed-4-bit scanners that handle odd `m`, but generic IVF-PQ does not. Could we either retain the even-`m` requirement for IVF-PQ 4-bit indexes or update its writer, reader, and both scan layouts consistently?

include/paimon_vindex.hpp
```diff
@@ -328,11 +360,24 @@ class Writer {
...
         raw.ctx = input_.get();
         raw.read_ranges_fn = detail::input_read_ranges;
-        handle_ = paimon_vindex_reader_open(raw);
+        raw.preferred_alignment_bytes = input_->preferred_alignment_bytes;
+        raw.preferred_window_bytes = input_->preferred_window_bytes;
+        raw.max_ranges_per_read = input_->max_ranges_per_read;
+        PaimonVindexReaderOptions options;
+        options.storage_profile = static_cast<uint32_t>(profile);
+        options.memory_budget_bytes = memory_budget_bytes;
+        handle_ = paimon_vindex_reader_open_with_options(raw, options);
```
leaves12138, line 470: Could we serialize native-handle operations in the C++ `Reader` as the Java wrapper now does (and cover destruction/move assignment as well)? The new DiskANN reader mutates lazy initialization, caches, workers, calibrated width, and search statistics during search, while each C entry point converts the same raw handle into a Rust `&mut VectorIndexReader`. Two C++ threads calling `search` on one `Reader` therefore create simultaneous mutable Rust references and can corrupt the reader; I reproduced that failure through the Python wrapper, which reaches the same C ABI. The callback thread-safety note only makes concurrent callbacks from one search safe; it does not make concurrent calls on the reader handle safe. If synchronization belongs to callers instead, the C and C++ APIs need an explicit non-thread-safe contract so users do not reasonably share this read-only-looking object across search threads.

jni/src/stream.rs
```diff
@@ -79,6 +80,28 @@ impl SeekRead for JniSeekableStream {
...
+        Ok(Some(self.clone()))
+    }
+
+    fn read_capabilities(&self) -> SeekReadCapabilities {
+        let Ok(mut env) = self.jvm.attach_current_thread() else {
+            return SeekReadCapabilities::default();
+        };
+        let read_hint = |env: &mut jni::JNIEnv<'_>, name: &str| -> usize {
+            env.call_method(self.stream_ref.as_obj(), name, "()J", &[])
+                .ok()
```
leaves12138: Could we avoid swallowing `call_method` failures here? If a custom `VectorIndexInput` implementation throws from one of these capability methods, `jni` returns `Error::JavaException` but the Java exception remains pending. This code converts the failure to `0` and continues opening the native reader; when the native method returns, the pending exception is propagated and the newly allocated reader handle is unreachable/leaked. The subsequent hint calls also run with an exception already pending. Please either read/validate these hints before constructing the reader and propagate the error, or explicitly clear the exception if falling back to zero is intentional. Negative hint values should probably be rejected consistently as well rather than silently becoming zero.

python/paimon_vindex/__init__.py
```diff
@@ -221,7 +258,16 @@ def create(cls, options: Mapping[str, str]):
...
-        with cls(options) as trainer:
+        data = _float32_matrix(data, "data")
+        resolved_options = dict(options)
+        if resolved_options.get("dimension") in (None, "auto"):
+            resolved_options["dimension"] = str(data.shape[1])
+        if (
+            resolved_options.get("index.type", "").startswith("ivf_")
+            and resolved_options.get("nlist") in (None, "auto")
+        ):
+            resolved_options["expected-vector-count"] = str(data.shape[0])
```
leaves12138, line 381: Could we preserve an explicitly supplied `expected-vector-count` here (and in the Java convenience method)? `data` is the training sample, not necessarily the final indexed corpus. With `nlist=auto`, a caller can intentionally pass `expected-vector-count=1000000` while training on 10,000 sampled vectors, but this assignment silently replaces it with `10000`, changing the inferred `nlist` from 1024 to 128. Please only infer `expected-vector-count` when the option is absent; an explicit corpus-size hint should take precedence over the convenience fallback.

core/src/ivfflat_io.rs
leaves12138: Thanks for adding aggregate payload batching. One oversized list is still unbounded, though: `bounded_ivf_payload_batch_end` deliberately returns the first payload even when it is larger than 64 MiB, and each reader allocates that whole list before issuing the positional read. A one-list IVF-FLAT index, or a highly skewed list, can therefore still allocate `count * dimension * 4` bytes (potentially multiple GiB), so the original OOM case is only fixed for the sum of multiple lists. Could we stream a single large list in chunks, or otherwise reject/cap it, while keeping the progress guarantee?

core/src/diskann.rs
leaves12138: Thanks for budgeting the per-task PQ scratch and parallelism. The retained trainer reservoir is still outside this plan when downsampling is required. `VectorIndexTrainer` always keeps up to 50,000 raw vectors; at `d=1024` that is 204,800,000 bytes. With a 128 MiB budget, `pq_training_plan` succeeds with `sample_count=27,892`, after which `bounded_pq_training_sample` allocates another sampled buffer while the original 50,000-vector reservoir remains alive. Cosine training can add a normalized owned copy as well. Since `peak_for` counts only one `sample_bytes` buffer, the accepted budget can still be exceeded before KMeans scratch is considered. Could the trainer reservoir limit be derived from the budget, or could the plan include the retained, downsampled, and normalization buffers that coexist?

core/src/pq.rs
leaves12138: The persistence path now rejects odd `m`, but the public in-memory API is still incorrect. `IVFPQIndex::with_nbits(..., 4, ...)` accepts odd `m`; both `scan_4bit_simd` and the transposed/FastScan path use `m / 2`, so the final subquantizer is ignored. I reproduced this with `m=3` and two codes that differ only in the third subquantizer: `ProductQuantizer::distance_from_table` distinguishes them, while `IVFPQIndex::search` ranks them as equal, both before and after `build_search_structures()`. Please either reject odd 4-bit `m` at construction/configuration time or teach all scanners to process the final low nibble.

python/paimon_vindex/__init__.py
leaves12138: The cross-thread race is fixed, but `threading.RLock` leaves a reentrant use-after-free path. `search` holds the lock while native code invokes `input.pread_many`; if that callback calls `reader.close()` (or another native reader method), it runs on the same thread and re-acquires the RLock, freeing/re-entering the handle while the original Rust call still owns `&mut VectorIndexReader`. I reproduced this with an IVF-FLAT input whose first search-time `pread_many` calls `reader.close()`; the Python process segfaults with exit status 139. The Java wrapper already tracks the native-handle owner and rejects reentry. Could the Python wrappers similarly reject all reentrant native-handle operations rather than permitting them? The writer output callbacks have the same pattern.

python/paimon_vindex/__init__.py
```diff
@@ -401,15 +497,55 @@ def __del__(self):
...
+            "local_storage": StorageProfile.LOCAL_STORAGE,
+            "remote_storage": StorageProfile.REMOTE_STORAGE,
+            "object_store": StorageProfile.OBJECT_STORE,
+        }
+        try:
+            storage_profile = profile_names.get(storage_profile, storage_profile)
+            storage_profile = StorageProfile(storage_profile)
+        except ValueError as exc:
+            raise ValueError(f"invalid storage_profile: {storage_profile}") from exc
+        if memory_budget_bytes < 0:
```
leaves12138: Could we apply the same platform-width validation added for `SearchParams` here? This only rejects negative values, so on a 64-bit build `memory_budget_bytes = ctypes.c_size_t(-1).value + 1` is silently converted to zero by `PaimonVindexReaderOptions`. The non-negative read-capability hints and the `l_search` / calibration `top_k` arguments have the same unchecked `c_size_t` conversion. Using `operator.index` and validating against `SIZE_MAX` before constructing the ctypes values would keep caller errors from changing the requested behavior.

tools/convert_ann_benchmarks.py
```diff
@@ -0,0 +1,149 @@
...
+
+import h5py
+import numpy as np
+
+
+def write_fvecs(
+    dataset: h5py.Dataset,
+    path: Path,
+    batch_rows: int,
+    row_limit: int | None = None,
```
leaves12138, line 34: The package declares `requires-python = ">=3.9"`, but these `int | None` annotations require Python 3.10 when annotations are evaluated normally. On Python 3.9 the script fails during function definition before the CLI can run. Please use `Optional[int]` or add `from __future__ import annotations`, and ideally include the minimum supported Python version in the tool smoke test.

core/src/diskann.rs
```diff
@@ -0,0 +1,1039 @@
...
+        let plan = pq_training_plan(
```
leaves12138: The high-level trainer reservoir is now budgeted, but the public low-level `DiskAnnIndex::train` path still calls `pq_training_plan` with one sample buffer. For cosine training with downsampling, this method simultaneously retains the sampled buffer returned by `bounded_pq_training_sample` and the normalized owned buffer returned by `preprocess_vectors`. With `d=1024`, `m=256`, 50,000 vectors, and a 128 MiB budget, the current plan selects 28,111 samples while the same peak model with two coexisting sample buffers selects only 14,379. Could this path pass the metric-aware sample-buffer count as well, and return an error rather than silently falling back to a one-vector plan when the budget is infeasible?

core/src/diskann_io.rs
```diff
@@ -0,0 +1,5419 @@
...
+        let plan = self.read_plan();
+        VectorIndexReadPlan {
+            random_read_latency_nanos: u64::try_from(self.random_read_latency.as_nanos())
+                .unwrap_or(u64::MAX),
+            window_bytes: plan.window_bytes,
+            max_ranges_per_read: self.read_capabilities.max_ranges_per_pread,
+            graph_beam_width: plan.graph_beam_width,
+            filtered_graph_beam_width: plan.filtered_graph_beam_width,
+            adjacency_preload_bytes: self.options.adjacency_preload_bytes,
+            adjacency_cache_bytes: self.options.adjacency_cache_bytes,
```
leaves12138: `vector_read_plan` reports the originally resolved cache budgets, not the effective capacities after lazy state consumes memory. `ensure_row_id_order` can shrink both shared caches through `resize_shared_cache_budgets`, but `self.options` remains unchanged. I extended `diskann_row_id_order_reserves_budget_from_shared_caches`: after loading the row-ID order, the public plan still reported 32,768 cache bytes while the actual combined cache capacity was 10,060 bytes. Since this API is documented as the concrete read plan, could it report the current hot-adjacency size and shared-cache capacities instead of the initial desired values?

include/paimon_vindex.hpp
```diff
@@ -355,35 +407,74 @@ class Reader {
...
 
     void optimize_for_search() {
-        check(paimon_vindex_reader_optimize_for_search(handle_));
+        std::lock_guard<std::mutex> lock(native_handle_mutex_);
+        check(paimon_vindex_reader_optimize_for_search(require_open()));
+    }
+
+    void warmup_queries(
+            const float* queries, size_t query_count, size_t l_search = 0) {
+        std::lock_guard<std::mutex> lock(native_handle_mutex_);
```
leaves12138: Serializing the C++ handle fixes the cross-thread race, but a plain `std::mutex` turns callback reentry into a permanent deadlock. I reproduced this with an IVF-FLAT `InputFile::read_ranges_fn` that calls `reader.metadata()` during `reader.search()`; the callback blocks reacquiring `native_handle_mutex_`, and the process timed out after five seconds. Java and Python now explicitly track the owning thread and reject this pattern. Could the C++ wrapper do the same rather than blocking same-thread reentry?

core/src/ivfsq_io.rs
```diff
@@ -0,0 +1,1312 @@
...
+    let mut list_to_queries = vec![Vec::new(); reader.nlist];
+    for (query_index, list_ids) in all_probe_indices.iter().enumerate() {
+        for &list_id in list_ids {
+            list_to_queries[list_id].push(query_index);
+        }
+    }
+    let d = reader.d;
+    let metric = reader.metric;
+    let mut heaps = (0..nq).map(|_| TopKHeap::new(k)).collect::<Vec<_>>();
+    let mut stream_scratches = (0..nq)
```
leaves12138: Streaming the oversized list bounds the I/O payload, but batch search retains one chunk-sized distance array for every query. I added a temporary unit test that scanned a 65,536-row chunk with 64 `SqScanScratch` instances; after the loop they retained 16,777,216 bytes (`64 * 65,536 * sizeof(f32)`) rather than one reusable 256 KiB buffer. Production chunk sizes can be much larger, so a large query batch can still consume GiBs despite the 64 MiB payload bound. Could queries be processed in bounded groups, or use scratch that is released/reused instead of one retained buffer per query? The transposed 8-bit IVF-PQ oversized-list path has the same per-query distance-scratch shape.

jni/src/stream.rs
```diff
@@ -166,3 +223,30 @@ impl paimon_vindex_core::io::SeekWrite for JniOutputStream {
...
+    use super::*;
+
+    #[test]
+    fn jni_seekable_stream_is_cloneable_for_parallel_diskann_batch() {
+        fn assert_clone<T: Clone>() {}
+        assert_clone::<JniSeekableStream>();
+    }
+
+    #[test]
+    fn jni_range_calls_are_bounded_to_one_local_reference_frame() {
```
leaves12138, line 241: The input path now bounds local references, but the JNI output path still retains one `JByteArray` local reference per `write_all` call until the whole native `writeIndex` frame returns. I reproduced this with a one-vector IVF-FLAT index and `nlist=50,000`; under `-Xcheck:jni`, the local-reference count grew past 69,000 before the run timed out. The offset table alone triggers multiple writes per list. Could `write_all` use a bounded local frame or explicitly `delete_local_ref(jbuf)` after `call_method`?

cpp/test_vindex.cpp
```diff
@@ -143,48 +143,91 @@ static void run_roundtrip(
     writer.write_index(make_output(buf));
     ASSERT_TRUE(!buf.data.empty());
 
-    paimon::vindex::Reader reader(make_input(buf));
+    paimon::vindex::Reader* active_reader = nullptr;
+    bool reentrant_attempted = false;
+    bool reentrant_rejected = false;
+    auto input = make_input(buf);
+    if (expected_index_type == PAIMON_VINDEX_INDEX_TYPE_IVF_FLAT) {
+        auto base_read = input.read_ranges_fn;
```
leaves12138, line 153: `base_read` is local to this `if` block, but the replacement callback captures it by reference through `[&]` and is invoked only after the block has ended. That leaves a dangling `std::function` reference in `input.read_ranges_fn`. The current C++ CI aborts while opening the IVF-FLAT reader (`failed to open vector index reader`), while my local optimized build happens to pass, which is consistent with this use-after-scope. Could the callback capture `base_read` by value (for example, `base_read = std::move(base_read)`) instead?

include/paimon_vindex.hpp
```diff
@@ -54,11 +56,59 @@ struct OutputFile {
...
 
 namespace detail {
 
+class NativeHandleMutex {
+public:
+    void lock() {
+        const auto current = std::this_thread::get_id();
+        {
+            std::lock_guard<std::mutex> state_lock(state_mutex_);
+            if (owner_ == current) {
```
leaves12138: The owner-thread check still deadlocks when the storage callback is invoked by a native worker thread rather than by the thread that entered the Reader. I reproduced this with `RAYON_NUM_THREADS=4`, a DiskANN `search_batch`, and an 8 KiB Reader budget so adjacency reads remain cold. A worker-thread `read_ranges_fn` callback called `reader.metadata()`; the outer search held `operation_mutex_` on the caller thread and waited for the worker, while the worker saw a different `owner_` and blocked on the same mutex. The process timed out after five seconds. Since DiskANN explicitly invokes callbacks from separate query workers, could reentry be tracked for the active native operation/callback context rather than only by comparing thread IDs?

python/paimon_vindex/__init__.py
```diff
@@ -15,21 +15,70 @@
...
+
+class _NativeHandleLock:
+    """Serialize native handles while failing same-thread callback reentry."""
+
+    def __init__(self):
+        self._lock = threading.Lock()
+        self._local = threading.local()
+
+    def __enter__(self):
+        if getattr(self._local, "active", False):
```
leaves12138: The same cross-thread callback reentry deadlock exists in the Python wrapper. `_NativeHandleLock` uses thread-local state, so a Rayon worker callback sees `active == false` and then blocks on the lock held by the caller whose native batch search is waiting for that worker. I reproduced this with a 5,000-row DiskANN index, `RAYON_NUM_THREADS=4`, a resident budget only 8 KiB above the required 75,254 bytes, and a 64-query batch. The control run completed with 488 reads across the caller plus four worker threads; when the first worker-thread `pread_many` callback called `reader.metadata()`, the process timed out after ten seconds. Could callback reentry be rejected based on the active native operation/callback context rather than per-thread local state?

java/src/main/java/org/apache/paimon/index/vector/VectorIndexReader.java
```diff
@@ -77,6 +84,55 @@ public void optimizeForSearch() {
...
+                return VectorIndexNative.calibrateSearchWidth(
+                        requireOpen(), queries, queryCount, topK);
+            } finally {
+                exitNativeHandle();
+            }
+        }
+    }
+
+    public VectorIndexReadPlan readPlan() {
+        synchronized (nativeHandleLock) {
```
leaves12138, line 134: The owner-thread check has the same cross-thread callback deadlock as the C++ and Python wrappers. DiskANN batch search can invoke `VectorIndexInput.pread` on a Rayon worker while the caller thread holds `nativeHandleLock`; if that worker callback calls `reader.metadata()` (or another Reader method), it blocks entering this `synchronized` section, while the outer native search waits for the callback. I reproduced it with the current head using `RAYON_NUM_THREADS=4`, the 5,000-row DiskANN fixture, a resident budget only 8 KiB above the required state, and 64 queries: the worker printed its reentry attempt and the Java process timed out after 12 seconds. Could callback reentry be rejected based on the active native operation/callback context rather than only `Thread.currentThread()`? A regression should exercise a callback originating from a native worker, not only same-thread IVF callbacks.

leaves12138 (commented):
Re-reviewed the updated head `5a8c44f` in ten focused passes.

The five findings from the previous round are addressed: the low-level cosine PQ training plan is metric-aware and propagates infeasible budgets; `read_plan` reports effective cache capacities; oversized IVF batch scans reuse one distance scratch; JNI output writes use bounded local frames; and the C++ same-thread callback test no longer captures a dead local variable.

I found three remaining reentry deadlocks in the new native-handle serialization:

1. C++ `NativeHandleMutex` only rejects the thread that owns the outer operation. A DiskANN worker-thread callback sees a different thread ID and blocks on the mutex while the outer batch search waits for that callback.
2. Python `_NativeHandleLock` has the same issue because its active marker is thread-local. The current-head reproducer prints the worker reentry and times out after 10 seconds.
3. Java `nativeHandleOwner` also only tracks the caller thread. A worker `VectorIndexInput.pread` callback that calls `reader.metadata()` prints the worker reentry and times out after 12 seconds.

All three need regression coverage with callbacks originating from native DiskANN query workers, not only same-thread IVF callbacks.

Validation on the current head:
- Rust library/integration tests: core 392 passed / 1 ignored, support 7 passed, fixtures 2 passed / 1 ignored, FFI 4 passed, JNI 2 passed
- Python 3.9 benchmark-tool syntax check and Python tests: 25 passed
- Java API/JNI validation, panic-boundary, and handle-safety tests passed
- C: 8 checks passed
- C++: 5 roundtrips passed
- GitHub CI is green

I am leaving this as COMMENTED rather than approving because the cross-thread callback deadlocks remain reproducible.

cpp/test_vindex.cpp
```diff
@@ -208,52 +254,77 @@ static void test_supported_index_roundtrips() {
...
 }
 
+static void test_worker_callback_reentry_is_rejected() {
+    int callback_context = 0;
+    paimon::vindex::detail::NativeHandleMutex mutex;
+    mutex.set_callback_context(&callback_context);
+    std::atomic<bool> rejected(false);
+
+    std::lock_guard<paimon::vindex::detail::NativeHandleMutex> operation(mutex);
+    std::thread callback_worker([&]() {
```
leaves12138, line 310: This new `std::thread` dependency is not linked by `cpp/CMakeLists.txt`. The GitHub Ubuntu runner passes because newer glibc folds pthread symbols into libc, but the same documented CMake build fails on GCC 12 / glibc 2.32 with `undefined reference to pthread_create` and `DSO missing from command line`. Could we add `find_package(Threads REQUIRED)` and link `Threads::Threads` to `test_vindex_cpp`? I verified the rest of the current-head C++ code builds once `-pthread` is supplied.

leaves12138 (commented):
Re-reviewed head `3ca82b5`.

The worker-callback reentry deadlocks are fixed across C++, Python, and Java. I reran the original end-to-end DiskANN reproductions with four Rayon workers: same-handle reentry is rejected promptly, and the outer batch search completes when the callback handles that rejection. The new Java native-worker regression also passes.

I found one remaining portability issue in the added C++ test: it now uses `std::thread`, but `cpp/CMakeLists.txt` does not link `Threads::Threads`. The documented CMake build fails on GCC 12 / glibc 2.32 with an unresolved `pthread_create`; GitHub's newer Ubuntu/glibc runner hides this and stays green. The test and all C++ roundtrips pass when `-pthread` is supplied. I left an inline comment with the CMake fix.

Other validation passed: formatting, Clippy, source/license checks, Rust library and integration tests, Python 26 tests, Java/JNI tests, C tests, and all GitHub checks.

Leaving this as COMMENTED rather than approving until the portable pthread link is added.

leaves12138 (approved):
Re-reviewed the latest head. The explicit `Threads::Threads` linkage fixes the C++ test build on GCC 12 / glibc 2.32, where the previous head failed with an unresolved `pthread_create`. The original CMake command now builds and all six C++ checks pass without manual flags.

The worker-callback reentry fixes remain valid across C++, Python, and Java, all GitHub checks are green, and I found no additional issues in the final incremental review.

PR #65, opened 2026-07-28
ivfpq: Reuse list tables in batch search

shyjsarah:
## Summary

Improve 8-bit IVF-PQ residual L2 batch-search performance by reusing batch-local list tables when many queries probe the same lists. This reduces repeated distance-table construction across queries and applies to both filtered and unfiltered batch search.

The optimization exposes `off`, `on`, and `auto` runtime modes. The default is `auto`, which enables reuse only when the batch and probe pattern is expected to amortize the table setup cost. 4-bit indexes retain their existing scan path and do not use batch-local ephemeral table reuse.

## Changes

- Build the list-dependent part of residual L2 PQ tables once per active loaded list.
- Build the query-dependent inner-product table once per active query in each loaded-list batch.
- Combine the list and query components before scanning instead of rebuilding a full residual table for every query/list pair.
- Apply the optimization to filtered and unfiltered 8-bit batch search.
- Add `IvfPqBatchTableReuseMode` with:
  - `off`: always use the direct residual-table path and ignore resident or batch-local precomputed tables.
  - `on`: request 8-bit batch-local table reuse whenever the metric/layout prerequisites hold and the memory budget permits it.
  - `auto`: additionally require the batch-size and probe-reuse cost heuristic.
- Keep 4-bit `on` and `auto` searches on the existing direct/resident table paths so results do not change with batch size.
- Use stable f64 arithmetic for batch-local list/query table construction to avoid catastrophic cancellation for 8-bit indexes with large vector offsets; scan tables and scan kernels remain f32.
- Compute stable PQ norms lazily once per search and reuse them across loaded-list batches.
- Bound ephemeral table memory with checked arithmetic, including list tables, shared PQ norms, and parallel query scratch tables; fall back to direct calculation when the budget is exceeded.
- Default Rust, Java, JNI, and legacy FFI search parameters to `auto`.
- Add Java `VectorSearchParams.withIvfPqBatchTableReuse(...)` overloads for enum and string values.
- Keep ephemeral tables batch-local so they do not remain resident on the reader.

## Testing

- `cargo test --workspace`
  - core: 413 passed, 1 ignored
  - ANN support: 7 passed
  - fixtures: 2 passed, 1 ignored
  - FFI: 4 passed
  - JNI: 3 passed
- `mvn -f java/pom.xml test` (BUILD SUCCESS)
- Verified filtered and unfiltered 8-bit batch results against individual reader searches with floating-point tolerance.
- Covered `off`, forced `on`, automatic activation and fallback, resident-table bypass, memory-budget fallback, and the large-offset 8-bit stability case.
- Verified 4-bit `auto` and `on` skip ephemeral precomputation, and 4-bit `auto` exactly matches `off` results for large batches.

## Notes

- PR #64 has been merged and is now part of the `main` baseline.
- The performance benefit comes from amortizing list-table construction across queries when `nq × nprobe` is sufficiently large. `auto` retains the direct path for workloads where reuse is unlikely to pay off.
- Stable f64 arithmetic is limited to ephemeral table setup; the distance scan path remains unchanged.
- The Paimon-side `vector_search` option wiring is being kept in a separate local change and is not part of this PR.

### DIFF
--- a/core/src/index.rs
+++ b/core/src/index.rs
@@ -33,9 +33,10 @@ use crate::ivfflat_io::{
     search_batch_ivfflat_reader, search_batch_ivfflat_reader_roaring_filter, write_ivfflat_index,
     IVFFlatIndexReader, IVFFLAT_MAGIC,
 };
+pub use crate::ivfpq::IvfPqBatchTableReuseMode;
 use crate::ivfpq::{
-    search_batch_reader, search_batch_reader_roaring_filter, search_with_reader,
-    search_with_reader_roaring_filter, IVFPQIndex,
+    search_batch_reader_roaring_filter_with_reuse_mode, search_batch_reader_with_reuse_mode,
+    search_with_reader, search_with_reader_roaring_filter, IVFPQIndex,
 };
 use crate::ivfrq::IVFRQIndex;
 use crate::ivfrq_io::{
@@ -988,6 +989,7 @@ pub struct VectorSearchParams {
     pub top_k: usize,
     pub search_width: SearchWidth,
     pub width: usize,
+    pub ivfpq_batch_table_reuse: IvfPqBatchTableReuseMode,
 }
 
 impl VectorSearchParams {
@@ -996,6 +998,7 @@ impl VectorSearchParams {
             top_k,
             search_width: SearchWidth::IvfNProbe,
             width: nprobe,
+            ivfpq_batch_table_reuse: IvfPqBatchTableReuseMode::Auto,
         }
     }
 
@@ -1004,6 +1007,7 @@ impl VectorSearchParams {
             top_k,
             search_width: SearchWidth::DiskAnnLSearch,
             width: l_search,
+            ivfpq_batch_table_reuse: IvfPqBatchTableReuseMode::Auto,
         }
     }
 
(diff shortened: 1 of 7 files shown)


JingsongLi:
When temporary table reuse is enabled, there is no restriction that `pq.nbits == 8`; therefore, the default `auto` setting also enters the 4-bit path and passes the newly generated f64→f32 table to the 4-bit quantization scan for processing (`core/src/ivfpq.rs:1633`).
- This causes **4-bit Top-K to vary with batch size**: when `nq < 64`, the old path is taken, but when `nq >= 64`, reuse may be automatically enabled, resulting in behavior that no longer matches that of the `off` mode or the behavior prior to the merge.

shyjsarah:
Thanks for catching this. Fixed in `81e6170` by requiring `pq.nbits == 8` before enabling batch-local ephemeral table reuse. As a result, 4-bit `auto` and `on` keep the existing direct/resident table paths, without changing the 4-bit scan kernels.

Added regression coverage that verifies:

- 4-bit `auto` and `on` do not create ephemeral list tables.
- 4-bit `auto` exactly matches `off` IDs and distances at the automatic activation threshold.

The full Rust workspace, Clippy, formatting, and Java tests pass.

JingsongLi:
+1

PR #67, opened 2026-07-31
ivfpq: Reuse query distance tables across IVF lists

jerry-024:
## Summary

Reuse non-residual IVF-PQ query distance tables across probed lists. For inner-product and cosine search, a PQ distance table depends only on the processed query and PQ codebook, so rebuilding it for every IVF list repeats identical work.

Batch search now applies the existing `off`, `on`, and `auto` reuse modes to this path. The default `auto` mode enables reuse only when the 8-bit layout, batch size, probe reuse, and memory-budget checks indicate that table setup should be amortized. Reuse memory is capped by `ivfpq_batch_table_reuse_max_bytes` (512 MiB by default); when the required tables do not fit, search falls back to direct table construction.

## Usage

Existing callers do not need to change anything. Rust, Java, C, C++, and Python IVF batch search defaults to `auto` with a 512 MiB reuse budget, so eligible 8-bit inner-product and cosine searches enable reuse automatically.

Java:

```java
VectorSearchParams params = VectorSearchParams.ivf(10, 8);
VectorSearchBatchResult result = reader.searchBatch(queries, queryCount, params);
```

To select a mode and memory budget explicitly:

```java
VectorSearchParams params =
        VectorSearchParams.ivf(10, 8)
                .withIvfPqBatchTableReuse(IvfPqBatchTableReuseMode.ON)
                .withIvfPqBatchTableReuseMaxBytes(256L * 1024 * 1024);
```

Rust:

```rust
use paimon_vindex_core::index::{
    IvfPqBatchTableReuseMode, VectorSearchParams,
};

let params = VectorSearchParams::new(10, 8)
    .with_ivfpq_batch_table_reuse(IvfPqBatchTableReuseMode::Auto)
    .with_ivfpq_batch_table_reuse_max_bytes(256 * 1024 * 1024);
let (ids, distances) = reader.search_batch(&queries, query_count, params)?;
```

Mode selection:

- `auto` (default, recommended): reuse only when the supported layout, batch-size, probe-reuse, and memory-budget checks pass.
- `on`: request reuse whenever the supported 8-bit layout, multiple probes, and memory budget permit it; useful for known large batch workloads and A/B benchmarks.
- `off`: keep the direct per-list table path; useful for comparison or troubleshooting.

C, C++, and Python expose the same reuse mode and max-byte budget. The C ABI uses additive V2 batch entry points; existing symbols remain compatible and use the defaults.

The mode affects IVF-PQ batch search. Other index types ignore it, and no index rebuild is required.

## Changes

- Build one non-residual PQ distance table per query and share it across probed lists in the in-memory and lazy-reader search paths.
- Reuse the same table for oversized streamed IVF lists.
- Extend batch reuse-mode behavior for 8-bit inner-product and cosine indexes:
  - `off`: retain direct per-list table construction.
  - `on`: reuse query tables when multiple lists are probed and the memory budget permits it.
  - `auto`: additionally require at least 64 queries and a 2x non-empty-probe reuse heuristic.
- Bound non-residual and residual-L2 resident/batch-local reuse with checked arithmetic and the configurable budget (512 MiB by default), falling back to direct table construction when it is exceeded.
- Add regression coverage for table-build counts and exact result equality between `off`, `on`, batch, and scalar search.
- Add an `ivfpq_batch_reuse_bench` benchmark that reports wall and CPU time at p50, p90, p95, and p99.

## Benchmark

Synthetic 8-bit inner-product IVF-PQ index on local Apple arm64, 11 Rayon workers, 100 measured batches after warmup:

- `N=99,840`, `D=768`, `M=16`, `nlist=256`, `nprobe=8`, `nq=256`, `k=10`

| Mode | p50 | p90 | p95 | p99 | p50 total CPU |
|---|---:|---:|---:|---:|---:|
| `off` | 34.270 ms | 36.913 ms | 38.977 ms | 42.413 ms | 326.921 ms |
| `on` | 7.017 ms | 8.002 ms | 8.333 ms | 9.730 ms | 56.649 ms |
| `auto` | 7.024 ms | 7.766 ms | 8.187 ms | 9.154 ms | 56.945 ms |

At p50, forced reuse is 4.9x faster and reduces total CPU time by about 82.7%. The benchmark verifies that `off`, `on`, and `auto` return identical IDs and distances before collecting samples.

## Testing

- `cargo +1.97.0 test -p paimon-vindex-core` (429 passed, 2 ignored across unit and integration tests)
- `cargo +1.97.0 clippy -p paimon-vindex-core --all-targets -- -D warnings`
- `cargo +1.97.0 build -p paimon-vindex-core --all-targets`
- `cargo +1.97.0 fmt --all -- --check`
- `git diff --check`
- `cargo +1.97.0 bench -p paimon-vindex-core --bench ivfpq_batch_reuse_bench`

## Notes

- No breaking API or on-disk format changes. C batch configuration is provided through additive V2 entry points; legacy C symbols remain available.
- The configured reuse budget is enforced for both non-residual query tables and resident/batch-local residual-L2 reuse.
- The 512 MiB default is intentional so larger native batches can reuse tables; callers can lower the per-search cap when memory is constrained.
- For the target benchmark shape, batch-local query tables use about 4 MiB per reader.

### DIFF
--- a/c/test_vindex.c
+++ b/c/test_vindex.c
@@ -339,6 +339,18 @@ static void run_roundtrip(
     }
     assert_id_in_cluster(batch_ids[0], 0);
     assert_id_in_cluster(batch_ids[1], 1);
+    struct PaimonVindexSearchParamsV2 batch_params_v2 = {
+        .top_k = 1,
+        .search_width = batch_params.search_width,
+        .width = batch_params.width,
+        .ivfpq_batch_table_reuse = PAIMON_VINDEX_IVFPQ_BATCH_TABLE_REUSE_OFF,
+        .ivfpq_batch_table_reuse_max_bytes = 1};
+    if (paimon_vindex_reader_search_batch_v2(
+            reader, queries, 2, batch_params_v2, batch_ids, batch_distances, 2) != 0) {
+        fail_ffi("reader search batch v2 failed");
+    }
+    assert_id_in_cluster(batch_ids[0], 0);
+    assert_id_in_cluster(batch_ids[1], 1);
     paimon_vindex_reader_free(reader);
     free(buf.data);
     free(data);
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -53,3 +53,7 @@ harness = false
 [[bench]]
 name = "diskann_bench"
 harness = false
+
+[[bench]]
+name = "ivfpq_batch_reuse_bench"
+harness = false
--- a/core/benches/ivfpq_batch_reuse_bench.rs
+++ b/core/benches/ivfpq_batch_reuse_bench.rs
@@ -48,6 +48,7 @@ struct Sample {
     system: f64,
 }
 
+#[cfg(unix)]
 fn cpu_times() -> CpuTimes {
     let mut usage = std::mem::MaybeUninit::<libc::rusage>::uninit();
     let result = unsafe { libc::getrusage(libc::RUSAGE_SELF, usage.as_mut_ptr()) };
@@ -56,6 +57,14 @@ fn cpu_times() -> CpuTimes {
     CpuTimes {
         user: usage.ru_utime.tv_sec as f64 + usage.ru_utime.tv_usec as f64 / 1_000_000.0,
         system: usage.ru_stime.tv_sec as f64 + usage.ru_stime.tv_usec as f64 / 1_000_000.0,
+    }
+}
+
+#[cfg(not(unix))]
+fn cpu_times() -> CpuTimes {
+    CpuTimes {
+        user: 0.0,
+        system: 0.0,
     }
 }
 
--- a/core/benches/ivfpq_batch_reuse_bench.rs
+++ b/core/benches/ivfpq_batch_reuse_bench.rs
@@ -1,0 +1,156 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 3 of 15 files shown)


JingsongLi:
1. **The configured memory limit will be ignored**: The resident residual-L2 path in `core/src/ivfpq.rs:1965` unconditionally builds `all_ip_tables` and does not validate `reuse_max_bytes`. Even if the configured value is very small, calling `optimize_for_search()` may still allocate `nq * m * ksub * 4` bytes.
2. **The default behavior results in an 8-fold increase in memory consumption**: `core/src/ivfpq.rs:937` sets the default budget to 512 MiB and applies it to the existing residual-L2 temporary tables in `core/src/ivfpq.rs:2143`; the limit in the old implementation was 64 MiB. This contradicts the statement in the PR description that “Residual L2 paths are unchanged”; existing Auto queries may consume up to an additional 512 MiB.

shyjsarah:
Confirmed. My remaining feedback is:

1. The reuse memory budget should be configurable consistently across all supported language bindings, including C, C++, and Python.
2. The configured budget should also be enforced for the optimized resident-precomputed L2 path.

jerry-024:
Thanks for the review. I updated the PR description to match the intended semantics and removed the inconsistent “existing 64 MiB” and “Residual L2 paths are unchanged” statements.

Addressed in `bb06131`:

1. Enforce `ivfpq_batch_table_reuse_max_bytes` before resident residual-L2 query-table allocation, with fallback to direct table construction when the budget is insufficient.
2. Expose the reuse mode and max-byte budget consistently in C, C++, and Python through ABI-compatible additive V2 C batch entry points; existing C symbols remain unchanged.
3. Keep the default at 512 MiB intentionally. For example, `nq=16384`, `M=16`, and 8-bit PQ requires 256 MiB of reusable query tables. The same configurable cap now governs non-residual and residual-L2 reuse, and callers can lower it per search when memory is constrained.

The commit has been pushed and is now the PR head.

shyjsarah:
+1

JingsongLi (approved):
+1

PR #69, opened 2026-08-04
fix: satisfy collapsible else-if lint

jerry-024:
Collapse the nested else-if in DiskANN adjacency validation so Clippy passes with warnings denied.

Tests:
- cargo fmt --all -- --check
- cargo clippy -p paimon-vindex-core --lib --target x86_64-unknown-linux-gnu -- -D warnings


shyjsarah:
+1

PR #70, opened 2026-08-05
ffi: Add extensible C search parameters

shyjsarah:
## Summary

Add an append-only C search-parameter structure so future query tuning fields can be exposed without adding a new by-value ABI version for every field.

This also exposes `max_initial_filter_expansion_factor` to C and C++ while preserving the existing V1/V2 symbols and behavior.

## Changes

- Add pointer-based `PaimonVindexSearchParamsEx` with a caller-provided `struct_size`.
  - Missing fields from shorter caller structs use existing defaults.
  - Unknown trailing fields from newer caller structs are ignored.
- Add `*_ex` variants for single, filtered, batch, and filtered-batch C searches.
- Keep all existing by-value C search functions unchanged for binary compatibility.
- Route the C++ RAII reader through the extensible API and expose:
  - `max_initial_filter_expansion_factor`
  - IVF-PQ batch table reuse mode
  - IVF-PQ batch table reuse memory budget
- Document the append-only ABI contract and C/C++ usage.

## Testing

- `cargo clippy --all-targets --workspace --locked -- -D warnings`
- `cargo test --workspace --locked`
- C FFI roundtrip tests, including a shorter prefix struct passed to the new API
- C++ roundtrip tests and query tuning parameter forwarding
- `git diff --check`

## Notes

- Public API: additive.
- Existing C ABI symbols and by-value parameter structures are unchanged.
- New callers should initialize `struct_size` with `sizeof(PaimonVindexSearchParamsEx)`.
- Follow-up to the query tuning added in #67 and #68.


jerry-024 (approved):
+1

PR #72, opened 2026-08-10
ivf: Retry only incomplete queries in automatic batch search

shyjsarah:
## Purpose

Automatic IVF batch search currently doubles `nprobe` for the entire batch whenever any query returns fewer than the required number of results. A small number of difficult queries can therefore force already-complete queries to be searched repeatedly at every expansion round.

## Changes

- Track incomplete queries independently during automatic IVF batch expansion.
- Pack and retry only incomplete queries in later rounds.
- Scatter retried results back to their original positions in the batch output.
- Apply the behavior to filtered and unfiltered IVF Flat, IVF SQ, IVF PQ, and IVF RQ batch search paths.
- Keep explicit/fixed `nprobe` searches as one unchanged full-batch call.
- Preserve the no-copy fast path when the first automatic-search round completes the batch.

This change intentionally does **not** reuse previously scanned IVF lists. Retried queries still search `0..nprobe`; incremental new-list-only scanning is handled by the follow-up PR.

## Tests

- `cargo test --workspace`
- `mvn -f java/pom.xml test`
- JNI release build and native validation, panic-boundary, and handle-safety tests on macOS

### DIFF
--- a/core/src/index.rs
+++ b/core/src/index.rs
@@ -64,6 +64,7 @@ use std::io::{self, Cursor};
 /// dimensions, for example `m=32` at 128 dimensions and `m=240` at 960.
 pub const DEFAULT_PQ_CODE_RATIO: f64 = 0.0625;
 const PERSISTED_ROW_ID_ESTIMATE_BYTES: usize = 10;
+const MAX_IVF_BATCH_RETRY_BUFFER_BYTES: usize = 64 * 1024 * 1024;
 
 /// Resolve a concrete PQ subquantizer count from a target code/raw byte ratio.
 ///
@@ -1846,7 +1847,8 @@ impl<R: SeekRead> VectorIndexReader<R> {
                 let total_vectors = usize::try_from(reader.total_vectors)
                     .map_err(|_| invalid_input("negative IVF vector count"))?;
                 let nprobe = params.resolve_ivf_nprobe(reader.nlist, total_vectors, None)?;
-                progressive_ivf_batch_search(
+                let mut aggregate_stats = IVFRQSearchStats::default();
+                let result = progressive_ivf_batch_search(
                     params,
                     reader.nlist,
                     nprobe,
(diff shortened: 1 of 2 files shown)


core/src/index.rs
```diff
@@ -2020,21 +2028,97 @@ fn progressive_ivf_search(
...
+                for &query_index in &active_queries {
+                    let start = query_index * dimension;
+                    packed.extend_from_slice(&queries[start..start + dimension]);
+                }
+                packed
+            };
+            &packed_queries
+        };
+
+        let (round_ids, round_distances) = search(round_queries, active_queries.len(), nprobe)?;
```
jerry-024: **[major] Avoid retaining the full-batch and retry buffers at the same time**

At this call, `result_ids` and `result_distances` for the original full batch are still live. A partial retry also owns `packed_queries`, and `search` allocates another IDs/distances pair. Peak memory therefore grows by `O(active_queries * (dimension + top_k))` on top of the full result buffers. For 100k queries at dimension 768 and `top_k = 100`, that is roughly 427 MB of additional live data before search-internal heaps and probe metadata, which can turn automatic expansion into an OOM.

Please avoid keeping all three representations live together—for example, replace the previous result directly when all queries remain active, and process partial retries in bounded chunks or pass active-query indices.

core/src/index.rs
```diff
@@ -1843,18 +1846,19 @@ impl<R: SeekRead> VectorIndexReader<R> {
...
+                progressive_ivf_batch_search(
                     params,
                     reader.nlist,
                     nprobe,
+                    queries,
                     query_count,
                     params.top_k,
                     total_vectors,
-                    |nprobe| {
+                    |active_queries, active_query_count, nprobe| {
```
jerry-024, line 1859: **[minor] Preserve statistics for the logical batch operation**

Each automatic retry calls `search_batch_ivfrq_reader` with `active_query_count`. That function initializes its aggregate `query_count` from this subset and overwrites `reader.last_search_stats`, so after progressive expansion `ivfrq_search_stats()` describes only the final retry and omits work performed for queries completed in earlier rounds.

Please aggregate IVF-RQ statistics across the internal rounds and publish one record with the original batch query count. The roaring-filter IVF-RQ path needs the same treatment.

core/src/index.rs
shyjsarah: Fixed in ce1bad4. Full-batch retries now drop the previous result buffers before searching again. Partial retries are processed in chunks bounded by a 64 MiB query/result buffer budget, with a regression test covering chunked retry/scatter behavior.

core/src/index.rs
shyjsarah: Fixed in ce1bad4. IVF-RQ statistics are accumulated across all internal progressive rounds, and the published query_count is restored to the original logical batch size for both filtered and unfiltered paths. Added a filtered progressive-search regression test.

jerry-024 (approved):
+1

PR #73, opened 2026-08-10
ivf: incrementally scan expanded batch probe ranges

shyjsarah:
## Summary

This is a follow-up to #72. After #72 removes completed queries from later automatic IVF batch rounds, incomplete queries still rerun the full larger `nprobe` and rescan old lists. This PR makes those retries incremental: a `16 -> 32 -> 64` expansion scans `[16,32)` and then `[32,64)` instead of rescanning `[0,32)` and `[0,64)`.

## Changes

- pass half-open probe ranges through automatic batch expansion for IVF-FLAT, IVF-SQ, IVF-PQ, and IVF-RQ
- seed each expansion round with the query's accumulated Top-K so prior heap results are retained
- scan only newly added lists in filtered and unfiltered batch paths
- preserve #72's bounded retry buffers when packing active queries and their Top-K seeds
- preserve aggregate IVF-RQ search statistics across progressive rounds
- keep fixed-width/scalar searches, storage formats, public APIs, and the C ABI unchanged

Logging and JNI diagnostic bridging are intentionally excluded from this PR and handled separately by #75.

## Testing

  - 451 passed, 1 ignored
- focused coverage for:
  - half-open probe ranges across multiple expansion rounds
  - carrying accumulated Top-K seeds between rounds
  - bounding packed query and seed buffers during partial retries
  - preserving a valid row ID of `-1` via its finite distance
  - IVF-PQ range partitioning and seeded result equivalence with a one-shot search
  - aggregate IVF-RQ statistics across progressive rounds

## Notes

- Rebased onto `main` after #72 was merged.
- The PR now contains one functional commit only.
- Query preprocessing and coarse centroid selection are still recomputed for the smaller active-query set in each round; no cross-round public search-state API is introduced.

### DIFF
--- a/core/src/ivfpq.rs
+++ b/core/src/ivfpq.rs
@@ -1769,15 +1769,75 @@ pub fn search_batch_reader_with_reuse_mode_and_budget<R: SeekRead>(
     reuse_mode: IvfPqBatchTableReuseMode,
     reuse_max_bytes: usize,
 ) -> io::Result<(Vec<i64>, Vec<f32>)> {
-    search_batch_reader_filter_with_reuse_mode_and_budget(
+    search_batch_reader_with_reuse_mode_and_budget_range(
         reader,
         queries,
         nq,
         k,
+        0,
         nprobe,
+        &[],
+        &[],
+        reuse_mode,
+        reuse_max_bytes,
+    )
+}
+
+pub(crate) fn search_batch_reader_with_reuse_mode_and_budget_range<R: SeekRead>(
+    reader: &mut IVFPQIndexReader<R>,
+    queries: &[f32],
+    nq: usize,
+    k: usize,
+    probe_start: usize,
+    probe_end: usize,
+    seed_ids: &[i64],
+    seed_distances: &[f32],
+    reuse_mode: IvfPqBatchTableReuseMode,
+    reuse_max_bytes: usize,
+) -> io::Result<(Vec<i64>, Vec<f32>)> {
+    search_batch_reader_filter_with_reuse_mode_and_budget_range(
+        reader,
+        queries,
+        nq,
+        k,
+        probe_start,
+        probe_end,
+        seed_ids,
+        seed_distances,
         None,
         reuse_mode,
         reuse_max_bytes,
+    )
+}
+
+fn search_batch_reader_filter_with_reuse_mode_and_budget_range<R: SeekRead>(
+    reader: &mut IVFPQIndexReader<R>,
+    queries: &[f32],
+    nq: usize,
+    k: usize,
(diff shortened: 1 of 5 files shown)


shyjsarah:
Diagnostic follow-up:

- Confirmed the PR #73 IVF-PQ batch diagnostic path was executing in EMR Spark, including the filtered batch route across all 14 shards.
- Root cause of the missing messages: the EMR log collector did not capture writes made directly to the Rust/native stderr descriptor.
- Added a JNI bridge that captures the opt-in IVF-PQ diagnostic line on the search thread and mirrors the same line through Java `System.err`, while retaining the existing native stderr write.
- EMR smoke job `jr-375b9ff8d65f53f3` succeeded: 14/14 filtered batch searches emitted complete diagnostics and there were no native errors.
- The smoke showed `nq=100`, `k=3`, fixed `nprobe=16`, and `incomplete_queries=0` on all 14 shards. Payload read time dominated filter and PQ scan time.

Minimal validation after cleanup:

- focused diagnostic capture tests (3 passed)
- `mvn -f java/pom.xml test`

The final change is contained in `paimon-vector-index`; no Paimon source change is required.

jerry-024:
New: 2; still open: 0; fixed candidates: 0.

core/src/ivfpq.rs
```diff
@@ -2365,13 +2444,44 @@ pub fn search_batch_reader_roaring_filter_with_reuse_mode_and_budget<R: SeekRead
...
+        0,
         nprobe,
+        &[],
+        &[],
+        roaring_filter_bytes,
+        reuse_mode,
+        reuse_max_bytes,
+    )
+}
+
```
jerry-024, line 2461: **[MAJOR]** The roaring-filter batch-search wrapper does not compile.

`search_batch_reader_roaring_filter_with_reuse_mode_and_budget_range` expects `(k, probe_start, probe_end, seed_ids, seed_distances, ...)`, but this call passes `(k, 0, 0, nprobe, &[], &[], ...)`. That supplies 13 arguments and shifts `nprobe: usize` into the `seed_ids: &[i64]` position. Remove the extra `0` so the call passes `k, 0, nprobe, &[], &[], ...`.

shyjsarah: Thanks for checking this. I verified the call on the current PR head. It passes `k, 0, nprobe, &[], &[]`, where `0` is `probe_start` and `nprobe` is `probe_end`; there is no second `0`.

The argument count and types match the range helper signature. The targeted Roaring-filter batch test and the full CI also pass on this commit, so no code change is needed here.

jerry-024 (approved):
+1

PR #76, opened 2026-08-12
perf(ivfpq): speed up transposed PQ8 scans

jerry-024:
## Summary

Speed up transposed 8-bit IVF-PQ scans by exposing each 256-entry lookup table as a fixed-size array, allowing the compiler to eliminate bounds checks in the hot loop. Search results and the existing 12.5% sparse/dense crossover remain unchanged.

## Changes

- Add fixed-array fast paths for initializing and accumulating transposed PQ8 distances.
- Keep the slice-based fallback for non-PQ8 lookup tables.
- Cover scan results around the configured sparse/dense crossover.
- Add a deterministic filter-scan benchmark using the production query shape (`nq=64`, `nprobe=64`).

## Testing

- [x] `cargo fmt --all --check`
- [x] `cargo check -p paimon-vindex-core --tests --target x86_64-unknown-linux-gnu`
- [x] `cargo clippy -p paimon-vindex-core --benches --target x86_64-unknown-linux-gnu -- -D warnings`

Single-thread benchmark on the same SSH development host:

| Filter density | p50 |
| ---: | ---: |
| 6.25% | 541.739 ms |
| 12.5% | 907.507 ms |
| 18.75% | 1058.931 ms |
| 25% | 1040.456 ms |
| 50% | 1071.382 ms |
| 100% | 1066.809 ms |

Historical scalar results on the same host and shape were 1890.700 ms at 50% density and 1849.500 ms at 100%, directional improvements of 43.3% and 42.3%. These were not interleaved A/B runs; production impact still needs a Spark run on the target AMD executor.

## Notes

- No unsafe code, architecture-specific SIMD, index-format changes, or user-facing configuration are added.
- Recall, filtering semantics, and the number of PQ codes evaluated are unchanged.

### DIFF
--- a/core/benches/ivfpq_filter_scan_bench.rs
+++ b/core/benches/ivfpq_filter_scan_bench.rs
@@ -90,7 +90,7 @@ fn main() {
         1,
         "run with RAYON_NUM_THREADS=1"
     );
-    assert!(NQ >= 64, "keep production query-table reuse enabled");
+    const { assert!(NQ >= 64, "keep production query-table reuse enabled") };
 
     let mut rng = StdRng::seed_from_u64(42);
     let mut index = IVFPQIndex::new(D, NLIST, M, MetricType::InnerProduct, false);
--- a/core/benches/ivfpq_filter_scan_bench.rs
+++ b/core/benches/ivfpq_filter_scan_bench.rs
@@ -19,6 +19,7 @@ use paimon_vindex_core::distance::MetricType;
 use paimon_vindex_core::io::{write_index, IVFPQIndexReader, PosWriter};
 use paimon_vindex_core::ivfpq::{search_batch_reader_filter, IVFPQIndex, RowIdFilter};
 use rand::rngs::StdRng;
+use rand::seq::SliceRandom;
 use rand::{Rng, SeedableRng};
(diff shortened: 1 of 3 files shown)


core/benches/ivfpq_filter_scan_bench.rs
```diff
@@ -0,0 +1,152 @@
...
+
+fn median(samples: &mut [Duration]) -> Duration {
+    samples.sort_unstable();
+    samples[samples.len() / 2]
+}
+
+fn main() {
+    assert_eq!((NQ, NPROBE), (64, 64), "benchmark production query shape");
+    assert!(
+        std::env::var_os("PAIMON_VINDEX_LOG_IVFPQ_TIMING").is_none(),
```
shyjsarah: This guard checks a stale environment variable name. The exercised search path currently reads `PAIMON_VINDEX_LOG_IVFPQ_BATCH_TIMING` and `PAIMON_VINDEX_LOG_IVFPQ_BATCH_REUSE`, so enabling either diagnostic still passes this assertion and can add instrumentation/logging overhead to the measured samples. Could we assert that both current variables are unset before running the benchmark?

shyjsarah:
+1

PR #77, opened 2026-08-13
fix: bound IVF-Flat serialization allocations

jerry-024:
## Summary

Bound IVF-Flat serialization memory usage by streaming vector payloads through one reusable write buffer instead of materializing full sorted-vector and encoded-byte buffers for each list.

The serialized output remains byte-for-byte compatible with the existing IVFL v1 reader and is independent of write chunk boundaries.

## Changes

- Reuse a private 64 MiB write buffer for centroid and inverted-list vector encoding.
- Preserve row-ID sorting while encoding referenced vectors in bounded chunks.
- Split oversized vectors across writes so the reusable buffer and vector payload writes stay within 64 MiB.
- Add regression coverage for byte stability, ID/vector ordering, multi-chunk writes, and the write-size bound.

## Testing

- [x] `cargo +1.94.0 test -p paimon-vindex-core ivfflat_chunked_writer`
- [x] `cargo +1.94.0 test -p paimon-vindex-core --lib` (446 passed, 1 ignored)

## Notes

- No public API, dependency, index option, or file-format version changes.
- The index vectors, row-ID sort permutations, and encoded IDs remain resident; this change only removes the two full vector-payload-sized serialization allocations.

### DIFF
--- a/core/src/ivfflat_io.rs
+++ b/core/src/ivfflat_io.rs
@@ -38,14 +38,31 @@ pub const IVFFLAT_HEADER_SIZE: usize = 64;
 const FLAG_DELTA_IDS: u32 = 1 << 0;
 const REQUIRED_FLAGS: u32 = FLAG_DELTA_IDS;
 const SUPPORTED_FLAGS: u32 = REQUIRED_FLAGS;
+const IVFFLAT_WRITE_BUFFER_SIZE: usize = 64 * 1024 * 1024;
 // Raw-vector scan cost scales with both rows and dimension. Below this amount,
 // Rayon scheduling and list-local heap merging outweigh the saved CPU time.
 const PARALLEL_FLAT_SCAN_MIN_COMPONENTS: usize = 1024 * 1024;
 
 pub fn write_ivfflat_index(index: &IVFFlatIndex, out: &mut dyn SeekWrite) -> io::Result<()> {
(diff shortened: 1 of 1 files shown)


core/src/ivfflat_io.rs
```diff
@@ -38,14 +38,31 @@ pub const IVFFLAT_HEADER_SIZE: usize = 64;
...
     let d = index.d;
     let nlist = index.nlist;
     validate_index_shape(index)?;
+    let bytes_per_vector = d.checked_mul(size_of::<f32>()).ok_or_else(|| {
+        io::Error::new(
+            io::ErrorKind::InvalidInput,
+            "IVF-FLAT bytes per vector overflow",
+        )
+    })?;
+    let buffer_limit = buffer_limit.max(bytes_per_vector);
```
shyjsarah: **Minor:** Could we keep `buffer_limit` as a hard byte limit and allow chunk boundaries inside a vector?

`buffer_limit.max(bytes_per_vector)` means a vector wider than 64 MiB is still fully materialized in `write_buffer` and submitted in one write. This is an improvement over the previous implementation, but it leaves the advertised memory bound dependent on the dimension.

Since the file format is independent of write boundaries, a wide vector could be encoded and written in multiple chunks. The wide-vector test could then assert `max_write <= TEST_BUDGET` instead of explicitly accepting an over-budget write.

shyjsarah:
+1

PR #79, opened 2026-08-17
chore: prepare release 0.4.0

shyjsarah:
## Purpose

Prepare the repository for the Apache Paimon Vector Index 0.4.0 release line. This is a release-readiness change only; it does not create, sign, or publish release artifacts.

## Changes

- align ASF branch protection with the `release-0.4` branch naming convention
- harden the release branch/version helper scripts, including Cargo lockfile refresh and local crate dependency version updates
- add automated tests for the release helper scripts and run them in CI
- refresh generated Rust dependency reports from internal version 0.3.0 to 0.4.0
- document the completed 0.3.0 release and move the upcoming release section to 0.4.0

## Verification

- `python3.11 -m unittest discover -s tools/tests -p 'test_*.py' -v`
- `python3 tools/check_license_headers.py`
- dependency license checks/report generation and binary legal inventory
- `cargo deny check licenses`
- `cargo fmt --check`
- Rust workspace tests (core, storage fixtures, FFI, and JNI)
- release workspace build
- C and C++ tests
- Java/JNI Maven tests on JDK 8, bundled JNI JAR verification, and loader smoke test
- Python tests, wheel build/content verification, installation, and import smoke test
- `cargo publish -p paimon-vindex-core --dry-run`
- two source archives generated from the same commit are byte-for-byte identical and contain the required release files
- fork CI passed all jobs on commit `0fff6387ffd18cc6c751979d45b8abd4b27d42b5`

### DIFF
--- a/tools/update_branch_version.sh
+++ b/tools/update_branch_version.sh
@@ -67,8 +67,8 @@ find . -name 'Cargo.toml' -not -path '*/target/*' -type f \
 perl -pi -e 's#^version = "'$OLD_VERSION_CLEAN'"#version = "'$NEW_VERSION_CLEAN'"#' python/pyproject.toml
 
 # Refresh workspace package versions in Cargo.lock while preserving the
-# already locked third-party dependency set, and reject inconsistent manifests.
-cargo check --workspace
+# already locked third-party dependency set without compiling the workspace.
+cargo update --workspace
 
 git commit -am "Update version to $NEW_VERSION"
 
--- a/tools/update_branch_version.sh
+++ b/tools/update_branch_version.sh
@@ -25,6 +25,7 @@ MVN=${MVN:-mvn}
 # fail immediately
 set -o errexit
 set -o nounset
+set -o pipefail
 # print command before executing
 set -o xtrace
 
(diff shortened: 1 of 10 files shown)


tools/update_branch_version.sh
```diff
@@ -36,33 +37,38 @@ fi
...
+	-exec perl -pi -e 'if (/^paimon-vindex-/) { s#version = "'$OLD_VERSION_CLEAN'"#version = "'$NEW_VERSION_CLEAN'"# }' {} \;
+
+# Change the Python package version.
+perl -pi -e 's#^version = "'$OLD_VERSION_CLEAN'"#version = "'$NEW_VERSION_CLEAN'"#' python/pyproject.toml
 
-#change version in pyproject.toml
-perl -pi -e 's#^version = "'$OLD_VERSION'"#version = "'$NEW_VERSION_CLEAN'"#' python/pyproject.toml
+# Refresh workspace package versions in Cargo.lock while preserving the
+# already locked third-party dependency set, and reject inconsistent manifests.
+cargo check --workspace
```
jerry-024: `cargo check --workspace` compiles the workspace, but this block only needs to refresh workspace package versions in `Cargo.lock`. Cargo documents `cargo update --workspace` for this exact case and it keeps already-locked non-workspace packages unchanged. Could we use that instead to avoid compiling during release preparation?

https://doc.rust-lang.org/cargo/commands/cargo-update.html#update-options

shyjsarah: Good point. I replaced `cargo check --workspace` with `cargo update --workspace` so the script refreshes only workspace package versions in `Cargo.lock` without compiling the workspace. I also added a regression assertion that the update does not create a Cargo `target` directory. The existing test still verifies manifest, path dependency, and lockfile version consistency.

jerry-024 (approved):
+1

PR #82, opened 2026-08-21
perf: batch IVF-PQ encoding with blocked SGEMM

jerry-024:
## Summary

Speed up IVF-PQ index builds by using a blocked 8-bit PQ batch encoder in the `add` path. Rows are processed in bounded Rayon blocks; per block, each sub-quantizer's centroid distances are computed with one SGEMM call and the same `||q||² + ||c||² − 2q·c` identity used by the canonical per-vector encoder and the search-time distance tables. Because add and search use identical arithmetic, codes stay byte-identical to `ProductQuantizer::encode` and no add/search scoring inconsistency can arise.

This supersedes the earlier transposed-codebook SIMD kernel from previous revisions of this PR. That approach used a different accumulation order than search, which required numerical-stability certificates and exact fallbacks to keep add and search consistent. Using the same arithmetic on both sides removes that problem class; the current diff is limited to the encoder, its benchmark, and CI lint compatibility.

## Changes

- Add `encode_batch_blocked`: a blocked SGEMM batch encoder for 8-bit PQ codes, used only from `IVFPQIndex::add`.
- Process rows in bounded Rayon blocks (≤512 rows per block) with per-worker scratch buffers.
- Fall back to the existing per-vector path for 4-bit PQ, sub-quantizers narrower than four dimensions, batches too small to supply at least four rows per worker (with a 32-row floor), and non-finite codebooks.
- Add a reproducible Criterion benchmark for the real `IVFPQIndex::add` entry point.

## Performance

Reproducible add-path benchmark:

```bash
RAYON_NUM_THREADS=8 cargo bench -p paimon-vindex-core --bench ivfpq_add_bench -- --noplot
```

Configuration: L2, 8 Rayon workers, Criterion release build, 500 ms warmup, 10 samples over 2 seconds per case. Each revision was run three times; the table reports the median of the three Criterion p50 estimates.

Machine: Apple M3 Pro (11 cores, 36 GiB), macOS, `aarch64-apple-darwin`. Main is the PR base `5400ab5` with only the benchmark harness from `2107a6f` applied. The optimized measurements are from `6e8e1bb`; the latest head `ce4fa22` has identical 8-worker dispatch for every case in this table.

| Case | Main p50 | PR p50 | Speedup |
| --- | ---: | ---: | ---: |
| `d=768, m=192`, 1 row | 0.158 ms | 0.161 ms | 0.98x |
| `d=768, m=192`, 7 rows | 0.398 ms | 0.385 ms | 1.03x |
| `d=768, m=192`, 8 rows | 0.390 ms | 0.391 ms | 1.00x |
| `d=768, m=192`, 31 rows | 0.947 ms | 0.997 ms | 0.95x |
| `d=768, m=192`, 32 rows | 1.098 ms | 0.898 ms | 1.22x |
| `d=768, m=192`, 512 rows | 13.877 ms | 8.898 ms | 1.56x |
| `d=768, m=192`, 4096 rows | 108.29 ms | 61.93 ms | 1.75x |
| `d=32, m=8` (`dsub=4`), 32768 rows | 35.77 ms | 19.76 ms | 1.81x |
| `d=64, m=8` (`dsub=8`), 32768 rows | 48.92 ms | 22.26 ms | 2.20x |
| Residual, `nlist=4096`, 32 rows | 3.996 ms | 3.775 ms | 1.06x |

With 8 workers, batches below 32 rows keep the original per-vector path and measured between 0.95x and 1.03x. At and above the 32-row threshold, the blocked path improves p50 by 1.22–2.20x; the residual case gains less because coarse assignment remains unchanged.

<details>
<summary>Individual 8-worker p50 trials (ms)</summary>

| Case | Main trials | PR trials |
| --- | --- | --- |
| 1 row | 0.154 / 0.159 / 0.158 | 0.165 / 0.161 / 0.154 |
| 7 rows | 0.398 / 0.399 / 0.389 | 0.378 / 0.395 / 0.385 |
| 8 rows | 0.388 / 0.398 / 0.390 | 0.402 / 0.390 / 0.391 |
| 31 rows | 0.947 / 0.983 / 0.922 | 0.997 / 1.034 / 0.940 |
| 32 rows | 1.112 / 1.086 / 1.098 | 0.898 / 0.913 / 0.890 |
| 512 rows | 13.628 / 14.079 / 13.877 | 8.865 / 8.898 / 9.342 |
| 4096 rows | 108.19 / 110.87 / 108.29 | 61.93 / 67.33 / 60.74 |
| `dsub=4`, 32768 rows | 35.77 / 36.48 / 35.62 | 19.96 / 19.76 / 19.69 |
| `dsub=8`, 32768 rows | 50.07 / 48.92 / 47.27 | 21.39 / 22.85 / 22.26 |
| Residual, 32 rows | 4.232 / 3.996 / 3.733 | 3.775 / 3.713 / 3.792 |

</details>

### Wide-pool threshold validation

The latest head `ce4fa22` was also compared with main at 32 Rayon workers for the small-batch cases raised in review, using the same Criterion settings and three trials per revision.

| Case | Main p50 | PR p50 | Speedup |
| --- | ---: | ---: | ---: |
| `d=768, m=192`, 32 rows | 1.080 ms | 1.122 ms | 0.96x |
| `d=768, m=192`, 64 rows | 2.047 ms | 2.005 ms | 1.02x |

At 32 workers the SGEMM threshold is 128 rows, so 32- and 64-row batches retain the per-vector path. Their p50 remains within 3.9% of main instead of switching to tiny SGEMM calls.

<details>
<summary>Individual 32-worker p50 trials (ms)</summary>

| Case | Main trials | PR trials |
| --- | --- | --- |
| 32 rows | 1.080 / 1.157 / 1.067 | 1.070 / 1.147 / 1.122 |
| 64 rows | 2.047 / 2.114 / 1.994 | 2.069 / 1.949 / 2.005 |

</details>

## Testing

- [x] `cargo test -p paimon-vindex-core` (462 passed)
- [x] `cargo clippy --all-targets --workspace`
- [x] Blocked codes match the per-vector encoder byte-for-byte (uniform and non-uniform chunk dims, large offsets)
- [x] Output is invariant to the add batch boundary, including NaN centroids
- [x] Benchmark coverage below, at, and above the 8-worker and 32-worker thresholds

## Compatibility

- No index-format or query-path change.
- Public `ProductQuantizer::encode_batch` remains byte-stable for existing storage fixtures.
- Newly built IVF-PQ codes are byte-identical to the legacy per-vector path (same arithmetic, same tie handling).

### DIFF
--- a/core/src/ivfpq.rs
+++ b/core/src/ivfpq.rs
@@ -252,7 +252,7 @@ impl IVFPQIndex {
 
         let code_size = self.pq.code_size();
         let mut codes = vec![0u8; n * code_size];
-        self.pq.encode_batch(&to_encode, n, &mut codes);
+        self.pq.encode_batch_blocked(&to_encode, n, &mut codes);
 
         for i in 0..n {
             let list_id = assignments[i];
--- a/core/src/pq.rs
+++ b/core/src/pq.rs
@@ -379,7 +379,7 @@ impl ProductQuantizer {
     /// Blocked batch encode for the IVF-PQ add path.
     pub(crate) fn encode_batch_blocked(&self, data: &[f32], n: usize, codes: &mut [u8]) {
         if self.nbits == 8
-            && n >= ENCODE_SGEMM_MIN_ROWS
+            && n >= encode_sgemm_min_rows(rayon::current_num_threads())
             && (0..self.m).all(|sub| self.chunk_dim(sub) >= 4)
             && self.centroids.iter().all(|value| value.is_finite())
         {
@@ -614,7 +614,12 @@ fn argmin_code(distances: &[f32]) -> u8 {
 
 /// Row block for the batched SGEMM encode path.
 const MAX_ENCODE_BLOCK_ROWS: usize = 512;
+const MIN_ENCODE_BLOCK_ROWS: usize = 4;
 const ENCODE_SGEMM_MIN_ROWS: usize = 32;
+
+fn encode_sgemm_min_rows(workers: usize) -> usize {
+    ENCODE_SGEMM_MIN_ROWS.max(workers.max(1) * MIN_ENCODE_BLOCK_ROWS)
+}
 
 fn encode_block_rows(rows: usize, workers: usize) -> usize {
     rows.div_ceil(workers.max(1))
@@ -864,10 +869,11 @@ mod tests {
     }
 
     #[test]
-    fn test_encode_block_rows_uses_available_workers() {
+    fn test_encode_block_rows_avoids_tiny_sgemm_blocks() {
         assert_eq!(encode_block_rows(2730, 12), 228);
         assert_eq!(encode_block_rows(32768, 12), MAX_ENCODE_BLOCK_ROWS);
-        assert_eq!(encode_block_rows(32, 64), 1);
+        assert_eq!(encode_sgemm_min_rows(8), 32);
+        assert_eq!(encode_sgemm_min_rows(32), 128);
     }
 
     #[test]
--- a/core/src/pq.rs
+++ b/core/src/pq.rs
@@ -356,6 +356,12 @@ impl ProductQuantizer {
     }
 
     /// Encode multiple vectors in parallel.
+    ///
+    /// This is the byte-stable path: results are bit-identical to the
+    /// per-vector [`Self::encode`], which golden storage fixtures rely on
+    /// (DiskANN serializes these codes). IVF-PQ's add path uses
+    /// [`Self::encode_batch_blocked`] instead, which batches the same distance
+    /// calculation into larger SGEMM calls.
     pub fn encode_batch(&self, data: &[f32], n: usize, codes: &mut [u8]) {
         let d = self.d;
         let cs = self.code_size();
@@ -368,6 +374,85 @@ impl ProductQuantizer {
                 }
             },
         );
+    }
+
+    /// Blocked batch encode for the IVF-PQ add path.
+    pub(crate) fn encode_batch_blocked(&self, data: &[f32], n: usize, codes: &mut [u8]) {
+        if self.nbits == 8
+            && n >= ENCODE_SGEMM_MIN_ROWS
+            && (0..self.m).all(|sub| self.chunk_dim(sub) >= 4)
+            && self.centroids.iter().all(|value| value.is_finite())
+        {
+            self.encode_batch_8bit_sgemm(data, n, codes);
+            return;
+        }
+        self.encode_batch(data, n, codes);
+    }
+
+    fn encode_batch_8bit_sgemm(&self, data: &[f32], n: usize, codes: &mut [u8]) {
+        let d = self.d;
+        let m = self.m;
+        let ksub = self.ksub;
+        let cs = self.code_size();
+        debug_assert_eq!(cs, m);
+        debug_assert_eq!(ksub, 256);
+
+        let max_dsub = (0..m).map(|sub| self.chunk_dim(sub)).max().unwrap_or(0);
+        let computed_norms = self
+            .centroid_norms_cache
+            .is_empty()
+            .then(|| self.compute_centroid_norms());
+        let centroid_norms = computed_norms
+            .as_ref()
+            .unwrap_or(&self.centroid_norms_cache);
+        let block_rows = encode_block_rows(n, rayon::current_num_threads());
+        codes[..n * cs]
+            .par_chunks_mut(block_rows * cs)
+            .enumerate()
+            .for_each_init(
+                || {
+                    (
+                        vec![0.0f32; block_rows * max_dsub],
+                        vec![0.0f32; block_rows * ksub],
+                    )
+                },
+                |(queries, distances), (block_idx, block_codes)| {
+                    let row0 = block_idx * block_rows;
+                    let rows = block_rows.min(n - row0);
+                    let block_data = &data[row0 * d..(row0 + rows) * d];
+
+                    for sub in 0..m {
+                        let range = self.chunk_range(sub);
+                        let dsub = range.len();
+                        for r in 0..rows {
+                            queries[r * dsub..(r + 1) * dsub].copy_from_slice(
+                                &block_data[r * d + range.start..r * d + range.end],
+                            );
+                        }
+                        let c_base = self.centroid_chunk_base(sub);
+                        sgemm_a_bt(
+                            rows,
+                            ksub,
+                            dsub,
+                            1.0,
+                            &queries[..rows * dsub],
+                            &self.centroids[c_base..c_base + ksub * dsub],
+                            0.0,
+                            &mut distances[..rows * ksub],
+                        );
+                        for r in 0..rows {
+                            let q_norm = fvec_norm_l2sqr(&queries[r * dsub..(r + 1) * dsub]);
+                            let row_distances = &mut distances[r * ksub..(r + 1) * ksub];
+                            for j in 0..ksub {
+                                row_distances[j] = (q_norm + centroid_norms[sub * ksub + j]
+                                    - 2.0 * row_distances[j])
+                                    .max(0.0);
+                            }
+                            block_codes[r * cs + sub] = argmin_code(row_distances);
+                        }
+                    }
+                },
+            );
     }
 
     /// Decode PQ codes back to an approximate vector.
@@ -527,6 +612,15 @@ fn argmin_code(distances: &[f32]) -> u8 {
     best as u8
 }
 
+/// Row block for the batched SGEMM encode path.
+const MAX_ENCODE_BLOCK_ROWS: usize = 512;
+const ENCODE_SGEMM_MIN_ROWS: usize = 32;
+
+fn encode_block_rows(rows: usize, workers: usize) -> usize {
+    rows.div_ceil(workers.max(1))
+        .clamp(1, MAX_ENCODE_BLOCK_ROWS)
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -663,6 +757,117 @@ mod tests {
         let table_distance = pq.distance_from_table(&table, &codes);
         let decoded_distance = fvec_l2sqr_sub(query, 0, &decoded, 0, d);
         assert!((table_distance - decoded_distance).abs() < 1e-4);
+    }
+
+    /// Reference per-vector encode used to pin the blocked batch path.
+    fn encode_per_vector(pq: &ProductQuantizer, data: &[f32], n: usize) -> Vec<u8> {
+        let cs = pq.code_size();
+        let mut codes = vec![0u8; n * cs];
+        for i in 0..n {
+            pq.encode(
+                &data[i * pq.d..(i + 1) * pq.d],
+                &mut codes[i * cs..(i + 1) * cs],
+            );
+        }
+        codes
+    }
+
+    #[test]
+    fn test_encode_batch_blocked_matches_per_vector() {
+        let d = 32;
+        let m = 8; // dsub = 4: hits the SIMD kernels
+        let mut rng = StdRng::seed_from_u64(20260820);
+        let train: Vec<f32> = (0..3000 * d).map(|_| rng.gen_range(-1.0f32..1.0)).collect();
+        let mut pq = ProductQuantizer::new(d, m);
+        pq.train(&train, 3000);
+
+        // Below, exactly at, above, and misaligned against the block size.
+        for n in [
+            1,
+            31,
+            32,
+            33,
+            MAX_ENCODE_BLOCK_ROWS,
+            MAX_ENCODE_BLOCK_ROWS + 7,
+            2048,
+        ] {
+            let data: Vec<f32> = (0..n * d).map(|_| rng.gen_range(-1.0f32..1.0)).collect();
+            let reference = encode_per_vector(&pq, &data, n);
+            let mut batch = vec![0u8; n * pq.code_size()];
+            pq.encode_batch_blocked(&data, n, &mut batch);
+            assert_eq!(batch, reference, "n={n}");
+        }
+    }
+
+    #[test]
+    fn test_encode_batch_blocked_non_uniform_chunks() {
+        let d = 13;
+        let m = 3;
+        let mut rng = StdRng::seed_from_u64(20260821);
+        let train: Vec<f32> = (0..2000 * d).map(|_| rng.gen_range(-1.0f32..1.0)).collect();
+        let mut pq = ProductQuantizer::with_nbits_balanced(d, m, 8);
+        pq.train(&train, 2000);
+        assert_eq!(pq.chunk_offsets, vec![0, 5, 9, 13]);
+
+        let n = MAX_ENCODE_BLOCK_ROWS + 13;
+        let data: Vec<f32> = (0..n * d).map(|_| rng.gen_range(-1.0f32..1.0)).collect();
+        let reference = encode_per_vector(&pq, &data, n);
+        let mut batch = vec![0u8; n * pq.code_size()];
+        pq.encode_batch_blocked(&data, n, &mut batch);
+        assert_eq!(batch, reference);
+    }
+
+    #[test]
+    fn test_encode_batch_blocked_matches_canonical_large_offset() {
+        let mut pq = ProductQuantizer::new(4, 1);
+        pq.centroids = vec![100_000_016.0; pq.d * pq.ksub];
+        pq.centroids[0..4].fill(100_000_008.0);
+        pq.centroids[4..8].fill(100_000_000.0);
+
+        let data = vec![100_000_000.0; 32 * pq.d];
+        let mut canonical = vec![0; 32];
+        pq.encode_batch(&data, 32, &mut canonical);
+        let mut blocked = vec![0; 32];
+        pq.encode_batch_blocked(&data, 32, &mut blocked);
+
+        assert_eq!(blocked, canonical);
+    }
+
+    #[test]
+    fn test_encode_batch_blocked_is_batch_invariant() {
+        let mut pq = ProductQuantizer::new(4, 1);
+        pq.centroids = vec![100.0; pq.d * pq.ksub];
+        pq.centroids[0..4].copy_from_slice(&[0.3658799, 0.06077051, -0.46501994, -0.31766486]);
+        pq.centroids[4..8].copy_from_slice(&[0.3658799, 0.06077051, -0.46501994, -0.31766483]);
+
+        let mut small = vec![0; 31];
+        pq.encode_batch_blocked(&[0.0; 31 * 4], 31, &mut small);
+        let mut large = vec![0; 32];
+        pq.encode_batch_blocked(&[0.0; 32 * 4], 32, &mut large);
+
+        assert_eq!(small.as_slice(), &large[..31]);
+    }
+
+    #[test]
+    fn test_encode_batch_blocked_nan_centroid_is_batch_invariant() {
+        let mut pq = ProductQuantizer::new(4, 1);
+        pq.centroids = vec![1.0; pq.d * pq.ksub];
+        pq.centroids[0] = f32::NAN;
+        pq.centroids[4..8].fill(0.0);
+
+        let mut small = vec![0; 1];
+        pq.encode_batch_blocked(&[0.0; 4], 1, &mut small);
+        let mut large = vec![0; 32];
+        pq.encode_batch_blocked(&[0.0; 32 * 4], 32, &mut large);
+
+        assert_eq!(small[0], large[0]);
+    }
+
+    #[test]
+    fn test_encode_block_rows_uses_available_workers() {
+        assert_eq!(encode_block_rows(2730, 12), 228);
+        assert_eq!(encode_block_rows(32768, 12), MAX_ENCODE_BLOCK_ROWS);
+        assert_eq!(encode_block_rows(32, 64), 1);
     }
 
     #[test]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: Clippy
-        run: cargo clippy --all-targets --workspace -- -D warnings
+        run: cargo clippy --all-targets --workspace -- -D warnings -A clippy::chunks-exact-to-as-chunks
 
   rust-test:
     runs-on: ubuntu-latest
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -65,3 +65,7 @@ harness = false
 [[bench]]
 name = "ivfpq_train_bench"
 harness = false
+
+[[bench]]
+name = "ivfpq_add_bench"
+harness = false
--- a/core/benches/ivfpq_add_bench.rs
+++ b/core/benches/ivfpq_add_bench.rs
@@ -1,0 +1,176 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use criterion::{
+    black_box, criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion, Throughput,
+};
+use paimon_vindex_core::distance::MetricType;
+use paimon_vindex_core::ivfpq::IVFPQIndex;
+use rand::rngs::StdRng;
+use rand::{Rng, SeedableRng};
+use std::time::Duration;
+
+#[derive(Clone, Copy)]
+struct Case {
+    name: &'static str,
+    d: usize,
+    m: usize,
+    nlist: usize,
+    rows: usize,
+}
+
+const CASES: [Case; 10] = [
+    Case {
+        name: "dsub4",
+        d: 768,
+        m: 192,
+        nlist: 1,
+        rows: 1,
+    },
+    Case {
+        name: "dsub4",
+        d: 768,
+        m: 192,
+        nlist: 1,
+        rows: 7,
+    },
+    Case {
+        name: "dsub4",
+        d: 768,
+        m: 192,
+        nlist: 1,
+        rows: 8,
+    },
+    Case {
+        name: "dsub4",
+        d: 768,
+        m: 192,
+        nlist: 1,
+        rows: 31,
+    },
+    Case {
+        name: "dsub4",
+        d: 768,
+        m: 192,
+        nlist: 1,
+        rows: 32,
+    },
+    Case {
+        name: "dsub4",
+        d: 768,
+        m: 192,
+        nlist: 1,
+        rows: 512,
+    },
+    Case {
+        name: "dsub4",
+        d: 768,
+        m: 192,
+        nlist: 1,
+        rows: 4096,
+    },
+    Case {
+        name: "coverage_dsub4_batch32768",
+        d: 32,
+        m: 8,
+        nlist: 1,
+        rows: 32768,
+    },
+    Case {
+        name: "coverage_dsub8_batch32768",
+        d: 64,
+        m: 8,
+        nlist: 1,
+        rows: 32768,
+    },
+    Case {
+        name: "coverage_e2e_residual",
+        d: 768,
+        m: 192,
+        nlist: 4096,
+        rows: 32,
+    },
+];
+
+fn new_index(
+    case: Case,
+    quantizer_centroids: &[f32],
+    centroids: &[f32],
+    norms: &[f32],
+) -> IVFPQIndex {
+    let mut index = IVFPQIndex::new(case.d, case.nlist, case.m, MetricType::L2, false);
+    index.quantizer_centroids = quantizer_centroids.to_vec();
+    index.pq.centroids = centroids.to_vec();
+    index.pq.centroid_norms_cache = norms.to_vec();
+    index
+}
+
+fn bench_ivfpq_add(c: &mut Criterion) {
+    let mut group = c.benchmark_group(format!("ivfpq_add/threads{}", rayon::current_num_threads()));
+    for case in CASES {
+        let mut rng = StdRng::seed_from_u64(42);
+        let data = (0..case.rows * case.d)
+            .map(|_| rng.gen_range(-1.0f32..1.0))
+            .collect::<Vec<_>>();
+        let ids = (0..case.rows as i64).collect::<Vec<_>>();
+        let quantizer_centroids = if case.nlist == 1 {
+            vec![0.0; case.d]
+        } else {
+            (0..case.nlist * case.d)
+                .map(|_| rng.gen_range(-1.0f32..1.0))
+                .collect()
+        };
+        let dsub = case.d / case.m;
+        let centroids = (0..case.m * 256 * dsub)
+            .map(|_| rng.gen_range(-1.0f32..1.0))
+            .collect::<Vec<_>>();
+        let norms = centroids
+            .chunks_exact(dsub)
+            .map(|centroid| centroid.iter().map(|value| value * value).sum())
+            .collect::<Vec<_>>();
+
+        group.throughput(Throughput::Elements(case.rows as u64));
+        group.bench_with_input(
+            BenchmarkId::new(
+                case.name,
+                format!(
+                    "d{}_m{}_dsub{dsub}_nlist{}_rows{}",
+                    case.d, case.m, case.nlist, case.rows
+                ),
+            ),
+            &case,
+            |b, &case| {
+                b.iter_batched(
+                    || new_index(case, &quantizer_centroids, &centroids, &norms),
+                    |mut index| index.add(black_box(&data), black_box(&ids), case.rows),
+                    BatchSize::LargeInput,
+                );
+            },
+        );
+    }
+    group.finish();
+}
+
+criterion_group! {
+    name = benches;
+    config = Criterion::default()
+        .sample_size(10)
+        .warm_up_time(Duration::from_millis(500))
+        .measurement_time(Duration::from_secs(2));
+    targets = bench_ivfpq_add
+}
+criterion_main!(benches);


shyjsarah (changes requested):
Thanks for the substantial performance work and validation. I found four issues that I think should be addressed before merge:

1. **The transposed PQ encoder can select a materially farther centroid** (`core/src/pq.rs`, around `encode_batch_8bit_transposed` / `score_argmin`).

   The new path evaluates `|c|² - 2q·c` directly in `f32`. With a large common component offset, cancellation can erase the distance difference. In a deterministic arm64 reproduction, the existing encoder selected code 0 with f64 squared L2 distance `0.0284440`, while the blocked encoder selected code 1 with distance `0.0319305` (~12% farther). This is not an ulp-level argmin tie, and the selected code is persisted by `IVFPQIndex::add`.

   Please use a numerically stable distance calculation (for example direct `(q-c)²`) or fall back to the stable path when the norm/dot formulation is ill-conditioned, and add this offset case as a regression test.

2. **Automatic approximate assignment can make a self-query miss at `nprobe=1`** (`core/src/ivfpq.rs`, `add_batch`).

   Build-time graph assignment stores the row in one approximate list, while query-time probing still chooses exact top-`nprobe` centroids. There is no guarantee that the assigned list is among those probed lists. In a deterministic out-of-sample reproduction, 127/2,000 rows were stored outside their exact nearest list; one row was stored in list 578 instead of 203, and querying the same vector with `k=1,nprobe=1` returned another row.

   Since this is automatically enabled for large shapes, it changes the recall behavior of existing no-option configurations. I suggest keeping exact assignment as the compatibility default, or exact-refining graph candidates / otherwise guaranteeing compatibility with query probing. Please also add end-to-end tests for self-query recall at low `nprobe`.

3. **The automatic cutoff does not account for graph-build amortization** (`core/src/ivfpq.rs`, `use_approximate_assignment`, `from_trained`, and `rebuild_assign_graph`).

   Every `from_trained` writer builds its own graph synchronously on first `add`. For a 2,730-row batch at `d=768,nlist=4096,m=192`, exact add measured roughly 36–42 ms, approximate first add roughly 89 ms, and approximate steady-state roughly 14 ms; it takes about four batches to recover the graph-build cost. Distributed writers processing only one or a few batches therefore regress.

   Please incorporate expected row count into the policy, delay construction until the break-even point, or build/share one immutable graph across writers.

4. **The fixed 512-row PQ block underutilizes Rayon workers** (`core/src/pq.rs`, `ENCODE_BLOCK_ROWS`).

   A 2,730-row batch creates only six tasks, regardless of whether the pool has 12, 32, or 64 threads. In a 12-core measurement, changing the block to 256 rows improved median time from 1.308 ms to 1.145 ms (~12.5%). A block size derived from `n` and `rayon::current_num_threads()` would avoid this cap.

Two smaller issues:

- `set_approximate_assignment(false)` does not clear an already-built `assign_graph`, and `add_batch` continues using that graph.
- The 1 GiB `memory_budget_bytes` supplied to `VamanaGraph::build` is not actually enforced; consider checking `estimate_vamana_memory_bytes` before construction.

The build-only graph lifecycle, retraining invalidation, compatibility-conscious config changes, and added tests are all good. However, the numerical correctness and low-`nprobe` recall regressions are blockers for me.

jerry-024:
Thanks for calling out the low-`nprobe` recall trade-off. This PR intentionally follows Lance’s add-only centroid-assignment optimization: for sufficiently large centroid matrices, add uses a small graph to accelerate assignment, while query continues to select `nprobe` centroids exactly. As in Lance, this does not guarantee that an approximately assigned list is included when `nprobe` is very low.

We treat this as an explicit performance/recall trade-off rather than an exact-assignment guarantee. Users that require exact coarse assignment, especially with a low `nprobe`, can disable the optimization with `approximate-assignment=false`. The default auto policy only enables it above the same `dimension × nlist` threshold used by Lance.

shyjsarah (changes requested):
Thanks for the follow-up fixes. I re-ran the previous reproductions against `29d13e9`.

The PQ numerical-cancellation issue, fixed PQ block parallelism, and stale assignment graph after disabling are fixed. The graph-startup amortization and graph-memory estimate are substantially improved; I consider the remaining edge cases non-blocking. I also acknowledge the low-`nprobe` behavior as the explicitly chosen approximate-assignment recall/throughput trade-off.

I found one new blocking API issue in the error-handling change:

**`IVFPQIndex::add` now panics on a recoverable explicit graph-build failure** (`core/src/ivfpq.rs`, around `add` / `try_add`).

```rust
pub fn add(&mut self, data: &[f32], ids: &[i64], n: usize) {
    self.try_add(data, ids, n)
        .expect("explicit IVF-PQ approximate assignment graph build failed");
}

pub(crate) fn try_add(...) -> io::Result<()> { ... }
```

`IVFPQIndex`, `with_approximate_assignment(true)`, and `add` are public Rust APIs. If the explicit assignment graph exceeds the budget or Vamana construction fails, direct core callers have no public `Result`-returning path: `try_add` is crate-private, so the recoverable error becomes a panic (or a process abort under `panic=abort`). `VectorIndexWriter`, FFI, and JNI propagate the error correctly, but they do not cover direct `IVFPQIndex` users.

Please expose a public non-panicking API—at minimum make `try_add` public and document/deprecate the panicking compatibility wrapper, or change `add` to return `io::Result<()>` when API compatibility permits. A regression test should verify that an explicit budget/build failure is returned rather than panicking through the supported public path.

Validation on this head: all GitHub checks pass, and the local core library suite completed with 472 passed / 1 ignored.

core/src/pq.rs
```diff
@@ -370,6 +376,84 @@ impl ProductQuantizer {
...
+        // batches and for the 4-bit packed path, which keeps the original
+        // per-vector implementation.
+        if self.nbits == 8
+            && n >= ENCODE_TRANSPOSE_MIN_ROWS
+            && (0..self.m).all(|sub| self.chunk_dim(sub) >= 4)
+        {
+            self.encode_batch_8bit_transposed(data, n, codes);
+            return;
+        }
+        self.encode_batch(data, n, codes);
```
JingsongLi, line 389: [P2] Keep PQ scoring consistent across IVF-PQ batch boundaries

This fallback sends batches smaller than 32 rows back through `encode_batch`, whose norm/dot L2 calculation can lose the distance difference through cancellation, while larger batches use direct squared differences. Since `IVFPQIndex::try_add` chunks at 32,768 rows, adding 32,799 identical rows can encode the first 32,768 as code 1 and the final 31 as code 0 solely because they land in the tail batch. A deterministic `d=4, m=1` reproduction with `q=[100000000; 4]`, centroid 0 `[100000008; 4]`, and centroid 1 `[100000000; 4]` selects centroid 0 in the tail even though the f64 squared distances are 256 and 0. Please keep the transpose threshold as a layout optimization, but use the numerically stable direct-difference scoring rule for the small-batch path as well.

core/src/vamana.rs
```diff
@@ -711,47 +719,76 @@ impl VamanaGraph {
         query: &[f32],
         search_list_size: usize,
     ) -> Vec<ScoredNode> {
+        let mut scratch = self.search_scratch(search_list_size);
```
JingsongLi: [P2] Clamp the public search width before allocating scratch

`search_scratch` preallocates several buffers directly from `search_list_size`, even when the graph contains far fewer nodes. The previous implementation remained bounded by `adjacency.len()`. On a small graph, `greedy_search(..., usize::MAX)` now panics immediately with `capacity overflow`, and less extreme oversized widths can cause avoidable large allocations. Please clamp the effective width to the graph size (and handle an empty graph) before constructing the scratch state.

core/src/ivfpq.rs
```diff
@@ -208,8 +280,68 @@ impl IVFPQIndex {
...
+                self.approximate_assignment = false;
+            }
+        }
+        Ok(())
+    }
+
     /// Add vectors in batches (Faiss-style: batch assign → batch residual → batch encode).
     pub fn add(&mut self, data: &[f32], ids: &[i64], n: usize) {
+        self.try_add(data, ids, n)
+            .expect("explicit IVF-PQ approximate assignment graph build failed");
```
JingsongLi: [P2] Expose a fallible low-level add path

Explicit approximate assignment introduces recoverable graph-build errors, including the fixed memory-budget rejection, but the public low-level `add` API converts them into a panic while `try_add` remains crate-private. `VectorIndexWriter`, FFI, and JNI correctly propagate the same error, so direct `IVFPQIndex` callers are the only callers that cannot recover and may abort under `panic=abort`. Please expose a public `Result`-returning add path, or change `add` to return `io::Result<()>` if compatibility permits.

core/src/index.rs
```diff
@@ -1158,6 +1162,36 @@ pub struct VectorIndexTrainer {
...
+            Some(enabled) => Self::new_with_approximate_assignment(config, enabled),
+            None => Self::new(config),
+        }
+    }
+
+    pub fn new_with_approximate_assignment(
+        config: VectorIndexConfig,
+        enabled: bool,
+    ) -> io::Result<Self> {
+        let mut trainer = Self::new(config)?;
```
JingsongLi: [P2] Reject unsupported configs in the typed constructor

`from_options` rejects `approximate-assignment` for every non-IVF-PQ index, but this public typed constructor accepts Flat, SQ, RQ, or DiskANN configs and silently ignores `enabled`. A caller applying the setting to a runtime-selected config therefore receives `Ok` even though the requested behavior was not honored. Please validate `config.index_type() == IndexType::IvfPq` here as well and add a non-IVF-PQ regression test.

core/src/ivfpq.rs
```diff
@@ -218,19 +350,59 @@ impl IVFPQIndex {
...
+            )?;
             offset += batch_n;
         }
+        Ok(())
     }
 
-    fn add_batch(&mut self, data: &[f32], ids: &[i64], n: usize) {
+    fn add_batch(&mut self, data: &[f32], ids: &[i64], n: usize) -> io::Result<()> {
+        if self.approximate_assignment && self.assign_graph.is_none() {
+            self.auto_assignment_rows_seen = self.auto_assignment_rows_seen.saturating_add(n);
```
JingsongLi: [P2] Do not amortize graph construction against rows that cannot benefit

This counter includes rows that were already assigned exactly, but those rows cannot amortize a graph built later. For example, 16,383 rows added in earlier calls followed by one final row causes the complete centroid graph to be built for that one-row tail; the same total input in one batch uses the graph for every row. This makes build cost and persisted assignment layout depend heavily on caller batching. Please base activation on expected remaining/total work, require enough current work to repay startup, or share/buffer the graph across writers.

core/src/ivfpq.rs
```diff
@@ -114,6 +162,24 @@ impl IVFPQIndex {
             codes: vec![Vec::new(); nlist],
             precomputed_table: Vec::new(),
             fastscan_codes: Vec::new(),
+            assign_graph: None,
+            approximate_assignment: use_approximate_assignment(d, nlist),
```
JingsongLi: [P2] Document the new default and its compatibility escape hatch

This silently changes existing large IVF-PQ builds to approximate coarse assignment once `dimension * nlist >= 1_000_000` and the row threshold is reached, which can change persisted list placement and low-`nprobe` recall. The public IVF-PQ option table and usage examples do not mention this default or the new `approximate-assignment=false` override. Please document the threshold, true/false/default semantics, recall trade-off, and the fact that this is build-only state that is not serialized.

JingsongLi (commented):
Inline review comments from a focused correctness, resource, performance, and public-API pass.

jerry-024:
Replaced the graph-based coarse assignment with exact projected branch-and-bound. On the 10M × 768 embedding-like benchmark, the previous Vamana path agreed with exact assignment for only 58.3% of rows and took 124.1 s; projected B&B took 102.7 s and preserved 100% exact agreement. This removes the recall and batch-determinism trade-off while improving add throughput.

core/src/projected_assign.rs
```diff
@@ -0,0 +1,715 @@
...
+        // Project uncentered vectors. Translation cancels in x-c, while this
+        // avoids a second source of f32 rounding in the lower bound.
+        sgemm_a_bt(nlist, dp, d, 1.0, cents, &proj, 0.0, &mut cents_p);
+        let cents_p_norms: Vec<f64> = (0..nlist)
+            .map(|c| norm_l2sqr_f64(&cents_p[c * dp..(c + 1) * dp]))
+            .collect();
+        let cents_p_norms_sqrt = cents_p_norms.iter().map(|norm| norm.sqrt()).collect();
+        let cents_p_errors: Vec<f64> = cents
+            .chunks_exact(d)
+            .map(|c| projection_error(norm_upper(c), d, dp))
```
JingsongLi: [P2] Keep Auto translation-invariant

Auto decides that this projection will pay off from the centered variance above, but this block projects uncentered centroids and makes `cents_p_errors` proportional to the full `||c||`; line 234 does the same for rows. Adding a common offset leaves every L2 distance unchanged but can inflate those error radii until every lower bound becomes zero. In a release/8-thread repro (`nlist=512,d=96,dp=8,n=4096`, rank-8 centroids transformed as `1e8 + (v-3)*64`), Auto evaluated all 2,097,152 row/centroid pairs and took 7.32 ms versus 1.82 ms for `find_nearest_batch` (4.0x slower). Please make the projection/error calculation safely translation-invariant, or make Auto reject projections based on sampled pruning effectiveness, and cover this translated low-rank case with an evaluation-rate regression.

core/src/ivfpq.rs
```diff
@@ -228,9 +293,16 @@ impl IVFPQIndex {
 
         // L2/IP without OPQ borrows the caller's batch instead of copying it.
         let processed = self.preprocess_queries(data, n);
-        let assignments =
-            kmeans::find_nearest_batch(&processed, n, &self.quantizer_centroids, self.nlist, d);
-
+        // Both branches return the exact nearest centroid; the projection only
+        // prunes the scan (see `projected_assign`).
+        let assignments = match &self.coarse_projection {
```
JingsongLi: [P2] Invalidate the projection when centroids change

`quantizer_centroids` remains publicly mutable, so an existing low-level caller can replace centroids after `train()`. Before this change, `add` scanned the current field; this branch instead uses bounds cached from the old centroids and can prune the new nearest list. In a deterministic repro (`d=64,nlist=512`, Auto projection), replacing centroid 500 with `centroid[0] + 0.1` and then adding that vector assigns it to list 0 (direct L2 about 0.64) instead of list 500 (distance 0), and residual encoding also uses the wrong centroid. Please make centroid replacement invalidate/rebuild `coarse_projection` (or make the centroids immutable behind an updating API), and add a mutate-after-train regression.

core/src/ivfpq.rs
```diff
@@ -200,17 +240,73 @@ impl IVFPQIndex {
...
+        self.coarse_projection = projection.map(Arc::new);
+        if self.coarse_projection.is_some() {
+            self.coarse_projection_centroids = Some(Arc::from(self.quantizer_centroids.as_slice()));
+        }
     }
 
     /// Add vectors in batches (Faiss-style: batch assign → batch residual → batch encode).
     pub fn add(&mut self, data: &[f32], ids: &[i64], n: usize) {
         const BATCH_SIZE: usize = 32768;
+        if self.coarse_projection.is_some()
```
JingsongLi: [P2] Avoid rescanning every centroid on each add

When a projection is active, this equality check walks all `nlist * d` floats on every public `add`, even when the centroids have not changed. In an arm64 release probe with `nlist=4096,d=768`, comparing the equal snapshot took 1.66 ms per call, while an exact `find_nearest` for one row took 0.423 ms; repeated small adds therefore spend about 4x an entire exact coarse assignment just validating the cache, erasing the projected path’s intended speedup. Please make centroid mutation invalidate/version the projection in O(1) (for example through an encapsulated mutation API) instead of validating the whole matrix on each unchanged add, and cover repeated small batches.

core/src/ivfpq.rs
```diff
@@ -2834,6 +2937,18 @@ fn seed_heaps(heaps: &mut [TopKHeap], seed_ids: &[i64], seed_distances: &[f32],
 
 // --- Utilities ---
 
+/// Evenly strided rows of the training data for projection calibration.
+fn calibration_sample(data: &[f32], n: usize, d: usize) -> Vec<f32> {
+    let rows = crate::projected_assign::CALIBRATION_ROWS.min(n);
+    if rows == 0 || d == 0 {
+        return Vec::new();
+    }
+    let stride = n / rows;
```
JingsongLi: [P2] Sample across the full calibration input

`stride = n / rows` floors the ratio before selecting rows. For `n=4095` and `rows=2048`, it becomes 1, so this takes exactly rows 0..2047 and ignores the entire latter half; at `n=4096` it abruptly changes to every other row. Because this sample drives Auto’s projected-width and cost gate, ordered training data can make the decision from an unrepresentative prefix and enable a path that is more expensive for the remaining rows (or reject a useful one). Please compute each source index fractionally (for example `i * n / rows`, with checked arithmetic, or endpoint spacing) and test the 2049/4095 boundaries with order-separated calibration data.

core/src/projected_assign.rs
```diff
@@ -0,0 +1,1195 @@
...
+            .collect();
+        let total_variance = centered
+            .iter()
+            .map(|v| (*v as f64) * (*v as f64))
+            .sum::<f64>()
+            / nlist as f64;
+        if total_variance.is_nan() || total_variance <= 0.0 {
+            return None;
+        }
+
```
JingsongLi: [P2] Skip PCA when Auto has no calibration sample

`set_quantizer_centroids` is now the public path for manually loading a trained coarse quantizer, and a newly constructed index has the default Auto mode but no `calibration_sample`. This function only discovers that Auto cannot select a projection after `top_subspace` has already fitted the full PCA. In a release/8-thread probe with `nlist=4096,d=768`, the default setter spent 423.8 ms here and then returned `None`; the same setter with projected assignment disabled took about 1 µs. Please clamp/validate `calibration_rows` and return early for `!force && calibration_rows < MIN_CALIBRATION_ROWS` before computing the mean, centered matrix, or PCA, and cover the default manual-load path.

core/src/projected_assign.rs
```diff
@@ -0,0 +1,1195 @@
...
+        d: usize,
+        block: usize,
+        force: bool,
+        calibration: &[f32],
+        calibration_rows: usize,
+    ) -> Option<Self> {
+        let exact_elapsed = if force {
+            None
+        } else {
+            let started = Instant::now();
```
JingsongLi: [P2] Stabilize the measured Auto gate

The exact path and each candidate width are timed exactly once, in a fixed order. These short Rayon/BLAS measurements are sensitive to scheduler and frequency noise, yet the winner is fixed for every subsequent `add`. In a release/8-thread fixed-data probe (`nlist=512,d=96`), 15 identical Auto fits selected `d′=16` eight times, `d′=24` six times, and rejected the projection once. Nine-call interleaved held-out medians for accepted fits ranged from 77.0% to 99.3% of exact time, so a one-shot fit admitted a path with essentially none of the advertised 15% margin. Please warm both paths, repeat/interleave the measurements, and compare a robust statistic; add a repeated-fit test/benchmark that selects the same fastest width and preserves the Auto margin.

core/src/ivfpq.rs
```diff
@@ -228,9 +337,16 @@ impl IVFPQIndex {
...
-            kmeans::find_nearest_batch(&processed, n, &self.quantizer_centroids, self.nlist, d);
-
+        // Both branches return the exact nearest centroid; the projection only
+        // prunes the scan (see `projected_assign`).
+        let assignments = match &self.coarse_projection {
+            Some(projection) => {
+                projection.assign(&processed, n, &self.quantizer_centroids, self.nlist)
+            }
+            None => {
+                kmeans::find_nearest_batch(&processed, n, &self.quantizer_centroids, self.nlist, d)
```
shyjsarah: **[Major — correctness]** These two branches do not actually have the same “exact nearest centroid” contract. `projection.assign` exact-checks candidates using direct squared differences, while `find_nearest_batch` uses the f32 norm-minus-dot GEMM path; batch query probing uses that GEMM convention as well. On valid finite vectors translated around `1e8`, I reproduced Enabled assigning the true list while Disabled/query probing selected another list. An index built with Enabled then missed its own two inserted vectors for a batch query with `nprobe=1` (returned `[-1, -1]`). Please use one stable distance/tie contract for both build and query, or conservatively exact-refine the GEMM result. A regression should cover Enabled/Disabled/Auto, scalar and batch paths, and serialized readers.

core/src/projected_assign.rs
```diff
@@ -0,0 +1,1195 @@
...
+        let total_variance = centered
+            .iter()
+            .map(|v| (*v as f64) * (*v as f64))
+            .sum::<f64>()
+            / nlist as f64;
+        if total_variance.is_nan() || total_variance <= 0.0 {
+            return None;
+        }
+
+        let (basis, eigenvalues) = top_subspace(&centered, nlist, d, block);
```
shyjsarah: **[Major — resource usage]** Auto reaches `top_subspace` before checking whether projected assignment is profitable, and the fit currently has no explicit memory/work budget. For `d=nlist=8192`, Auto chooses `block=2728`; the key PCA buffers can exceed 1 GiB and the eight iterations are roughly `6e12` FLOPs. A small training input can still reach this because coarse k-means repeats rows when `n <= nlist`. Please estimate checked peak bytes/work before allocation, cap `block*nlist` and `block^2`, and skip/fail safely before entering PCA when the configured budget would be exceeded.

core/src/projected_assign.rs
```diff
@@ -0,0 +1,1195 @@
...
+    ) -> (Vec<usize>, usize) {
+        let d = self.d;
+        let dp = self.dp;
+        debug_assert_eq!(self.cents_p.len(), nlist * dp);
+        let mut out = vec![0usize; n];
+        let (block_rows, _) =
+            crate::kmeans::assignment_block_plan(n, dp, nlist, rayon::current_num_threads());
+        let block_rows = block_rows.min(MAX_BLOCK_ROWS);
+        let evaluations = out
+            .par_chunks_mut(block_rows)
```
shyjsarah: **[Major — resource usage]** `assignment_block_plan` returns a `parallel` flag specifically to enforce a serial fallback when the aggregate scratch budget would be exceeded, but this path discards it and always uses `par_chunks_mut`. Each worker independently allocates the `rows*nlist` score matrix, `nlist` bounds, and potentially `nlist` candidates. For `nlist=1,048,576` and 32 workers, score+bounds are about 384 MiB aggregate and worst-case candidates add about 512 MiB, bypassing the intended budget. Please honor `parallel=false` and include all per-worker scratch—not only the score matrix—in the planner.

shyjsarah (changes requested):
Thanks for the substantial rework. I re-reviewed the current projected branch-and-bound implementation. The earlier Vamana/PQ/public-add issues appear resolved, and all current CI checks pass. However, I found three remaining issues that should be fixed before merge: (1) build assignment and batch query probing do not share the same stable distance contract, which can cause an nprobe=1 self-query miss; (2) Auto performs potentially unbounded PCA work before its profitability gate; and (3) projected assignment ignores the block planner’s serial fallback and can multiply scratch memory by the Rayon worker count. Details and suggested fixes are inline.

core/src/kmeans.rs
```diff
@@ -747,10 +786,10 @@ pub fn find_topk_batch(
     d: usize,
     nprobe: usize,
 ) -> (Vec<Vec<usize>>, Vec<Vec<f32>>) {
-    let centroid_norms = (0..k)
-        .map(|c| fvec_norm_l2sqr(&centroids[c * d..(c + 1) * d]))
-        .collect::<Vec<_>>();
-    find_topk_batch_with_centroid_norms(queries, nq, centroids, &centroid_norms, k, d, nprobe)
+    (0..nq)
```
shyjsarah: **[Major — resource usage]** The stable batch probing fix now runs one `find_topk` allocation per active Rayon query. Each call allocates `Vec<(f32, usize)>` with capacity `nlist` (16 bytes per centroid on 64-bit targets), with no aggregate query scratch budget. At `nlist=1,048,576` and 32 active workers this can keep roughly 512 MiB of tuple buffers live. A bounded allocator reproduction at `k=262,144, nq=32, threads=32` measured about 109 MiB peak on this head versus 39 MiB before the change. Please keep the direct-distance contract but bound/reuse storage—for example an O(nprobe) heap per query, a capped per-worker buffer, or query blocking based on `workers * nlist * tuple_size`.

core/src/kmeans.rs
```diff
@@ -705,16 +735,10 @@ pub(crate) fn find_nearest_batch(
...
-        return Vec::new();
-    }
-    if n == 1 {
-        return vec![find_nearest(&data[..d], centroids, k, d)];
-    }
-
-    let mut assignments = vec![0usize; n];
-    assign_clusters_fast(data, n, d, centroids, k, &mut assignments, 0.0);
-    assignments
+    (0..n)
```
shyjsarah: **[Major — performance regression]** Replacing the blocked batch assignment with independent row scans fixes cancellation, but the four-centroid kernel does not recover the lost batch-level reuse. In an 8-worker release comparison at `d=768, nlist=4096, n=2048`, the previous blocked path had a 38.1 ms median versus 59.6 ms here (about 56% higher latency / 36% lower throughput). One-worker batch probing also regressed from 1.88 to 3.58 ms and from 5.47 to 8.58 ms on two tested shapes. Since this is a build-performance PR and the exact fallback is used whenever Auto is unavailable/rejected, please add a numerically stable blocked direct-distance kernel (with a small-batch crossover), or exact-refine an error-bounded GEMM candidate set.

core/src/projected_assign.rs
```diff
@@ -0,0 +1,1259 @@
...
+        let dp = self.dp;
+        debug_assert_eq!(self.cents_p.len(), nlist * dp);
+        let mut out = vec![0usize; n];
+        // Per worker: bounds (f64) and candidates (f64 + u32), plus projected
+        // rows and their f64 norms. The planner already counts the score matrix.
+        let fixed_scratch = nlist.saturating_mul(
+            (std::mem::size_of::<f64>() + std::mem::size_of::<(f64, u32)>())
+                .div_ceil(std::mem::size_of::<f32>()),
+        );
+        let row_scratch = dp.saturating_add(2);
```
shyjsarah: **[Minor — incomplete scratch accounting]** Honoring `parallel=false` fixes the original worker-amplification case, but `row_scratch = dp + 2` omits transient buffers held by `project_rows`: `rows*d` f64 centered data, `rows*dp` f64 output, and the f32 projection allocated while those buffers are still live. The actual row cost is approximately `2*d + 3*dp + 2` f32 units. For `d=8192, nlist=dp=512, threads=32`, the planner models about 16 MiB but can select a parallel plan whose projection phase reaches roughly 271 MiB. Please account for the projection and score phases explicitly (or move to reusable caller-owned buffers), and add an end-to-end planner regression rather than testing only the supplied estimates.

shyjsarah (changes requested):
Thanks for the follow-up fixes. I re-ran the previous reproductions against `4b0efa9`. The distance-contract/self-query issue is fixed across Enabled/Disabled/Auto and scalar/batch/serialized-reader paths. The Auto PCA resource gate is also correctly placed before the expensive allocations. The original ignored-serial-fallback case is fixed, although the scratch accounting remains incomplete (inline). I found two new blockers introduced by the stable exact-scan change: batch top-k can multiply an O(nlist) tuple buffer across Rayon workers, and bulk exact assignment loses the previous blocked batch reuse with a substantial throughput regression. The existing non-blocking Auto issues also remain: one-row adds are slower, width selection uses one noisy timing sample, and setup cost is not amortized against expected add volume.

core/src/kmeans.rs
```diff
@@ -767,39 +1066,204 @@ pub(crate) fn find_topk_batch_with_centroid_norms(
...
+            let dist = approximate.max(0.0);
+            if dist < best {
+                second = best;
+                best = dist;
+                best_idx = c;
+            } else if dist < second {
+                second = dist;
+            }
+        }
+        let upper = best as f64 + 2.0 * certified_error(e_gemm, best as f64, d);
```
shyjsarah: **[Major — correctness]** The certification error is relative-only and does not include an additive allowance for gradual underflow. This can make the band hundreds of times smaller than one f32 subnormal ULP and certify the wrong centroid. With a finite deterministic `d=1` fixture, scalar `find_topk` selects centroid 0, while an eight-row one-thread blocked batch certifies centroid 8 for every row; a separate `nprobe=3` fixture also returns a strictly farther centroid. End-to-end, Disabled and Enabled assign different lists, scalar and batch `nprobe=1` queries probe different lists, and the disagreement survives IVFPQ/IVFFlat serialization on both arm64 and x86_64. Please add architecture-independent absolute underflow/FTZ terms to the norm, GEMM, combine, and direct-distance certificate—or fail closed to the direct path whenever terms can enter the subnormal range—and add the exact bit-pattern regressions.

core/src/kmeans.rs
```diff
@@ -728,16 +989,51 @@ pub fn find_topk(
...
+    centroids: &[f32],
+    k: usize,
+    d: usize,
+    nprobe: usize,
+) -> (Vec<Vec<usize>>, Vec<Vec<f32>>) {
+    let search = |qi| find_topk(&queries[qi * d..(qi + 1) * d], centroids, k, d, nprobe);
+    if use_parallel_direct_topk(nq, k, nprobe, rayon::current_num_threads()) {
+        (0..nq).into_par_iter().map(search).unzip()
+    } else {
+        (0..nq).map(search).unzip()
```
shyjsarah: **[Major — performance regression]** Bounding memory by switching this entire batch to a sequential iterator creates a large latency cliff for small query batches on large-`nlist` indexes. Matching-checksum 8-thread release runs at `k=262144, nq=8, nprobe=2` regressed from 3.58 to 10.02 ms for `d=8` and from 3.05 to 15.13 ms for `d=32` (about 2.8x-5.0x). At `k=1048576, nq=32, nprobe=2, threads=32`, the same mechanism is about 3x slower. The memory cap is necessary, but please retain bounded concurrency—for example, O(`nprobe`) per-worker heaps, centroid tiling with capped reusable buffers, or limiting the active query count instead of serializing every query.

shyjsarah (changes requested):
Thanks for the next iteration. I re-ran the earlier checks against `ffa0b86`. The projected scratch accounting now includes the `project_rows` temporaries, the previous worker-multiplied top-k allocation is substantially bounded, the large-translation contract regression remains fixed, and the certified blocked kernel fixes the previously reported bulk exact-assignment slowdown on its cited shapes. I still found two merge-blocking issues: the relative-only certificate can false-prune finite subnormal values, and the scratch fallback serializes the whole query batch and creates multi-fold latency regressions for large `nlist`. Details are inline. Non-blocking issues also remain around candidate `Vec` capacity, stable-sort scratch, the rows-only blocked cutoff, near-exhaustive probes doing duplicate work, Auto small-batch/timing/amortization behavior, and the independently mutable IVFRQ centroid/norm cache.

shyjsarah:
Re-reviewed the latest head (`2f0ffa9`). The earlier subnormal-correctness and small-`nprobe` serialization issues appear fixed, but two blockers remain:

1. **Scalar high-`nprobe` regression is still unresolved.** `find_topk` selects the bounded-heap path using `rayon::current_num_threads()`, even though a single scalar query is not parallelized internally. On the latest head, for `d=8, k=262144, nprobe=65536`, I measured a median of **2.51 ms** in a 1-thread pool versus **11.35 ms** in an 8-thread pool (~4.5x slower). The algorithm choice should not depend on ambient pool width for a scalar call; it should reflect actual concurrent queries/scratch usage, or use the linear-selection path here.

2. **CI is currently red.** Besides the formatting failure, `test_mixed_block_sgemm_only_processes_safe_rows` now gets `gemm_rows=0` instead of `16`. The test uses `d=1`, so the new `MIN_GEMM_DIM=32` cutoff routes it through the direct path and the test no longer exercises the mixed SGEMM branch it is intended to verify. Please update the fixture to use `d >= 32` while retaining mixed safe/unsafe rows.

So this is improved, but I don't think the PR is ready to merge yet.

shyjsarah:
Re-reviewed the latest head (`540d672`). The previous blockers are fixed: scalar `find_topk` is now independent of the ambient Rayon pool width, the mixed-SGEMM test was corrected, and all CI checks are green. The latest PQ batch-size stability change also looks correct.

I found one remaining performance blocker in the new batch worker policy (`core/src/kmeans.rs:96-109`, `1320-1336`):

**Batch-size threshold causes a large top-k latency cliff.** `workers = min(nq, threads)` is also used to decide between linear selection and the bounded heap. With:

```text
d=8, k=262144, nprobe=65536, Rayon threads=32
nq=4:  median 2.69 ms
nq=5:  median 12.67 ms
```

Adding one query makes the batch about **4.7x slower**. At four workers the aggregate linear-selection scratch is exactly within `MAX_MATRIX_ELEMS`; at five workers it crosses the limit and every query switches to the much slower heap path.

Please consider capping the number of concurrently executing linear-selection workers and processing excess queries in another wave, rather than changing the per-query algorithm based on batch width. This would keep the scratch bound without introducing the cliff.

Minor follow-up: the scratch estimates still do not include stable-sort temporary allocation, so the documented 16 MiB scratch bound can be exceeded for large/full probes.

So correctness and CI now look good, but I recommend addressing the batch performance cliff before merge.

shyjsarah:
Follow-up on the latest head (`b00138f`): the wave approach fixes the previously reported high-`nprobe` batch cliff, and replacing stable sorting also fixes the unbudgeted sort scratch. However, applying linear waves to every `nprobe > 1` introduces a larger regression for small probes.

At `core/src/kmeans.rs:96-102` and `1333-1354`, any configuration that can fit at least one linear worker enters the wave path, bypassing the parallel bounded-heap path even when `nprobe` is tiny. Reproduction:

```text
nq=8, d=8, Rayon threads=32

k=1,048,576, nprobe=1:  1.57 ms
k=1,048,576, nprobe=2: 21.74 ms
k=1,048,577, nprobe=2:  2.26 ms
```

With `k=1,048,576` and `nprobe=2`, `linear_workers=1`, so all eight queries run as serial linear waves. Increasing `k` by one makes the worker limit zero and falls back to the parallel bounded heap, making it roughly **9.6x faster**. `nprobe=1 -> 2` at the boundary is about a **13.8x latency cliff**.

This regression primarily affects **batch query coarse search**, not the IVF-PQ build/add path: construction assigns one nearest centroid through projected assignment or `find_nearest_batch`, while `find_topk_batch` with `nprobe > 1` is used by the IVF-PQ/IVF-Flat/IVF-SQ/IVF-RQ search paths.

I suggest enabling linear waves only when `nprobe` is large enough that linear selection is actually preferable; small/moderate `nprobe` should retain the parallel bounded-heap path. All CI checks are currently green and I did not find a correctness issue, but this query-path regression should still be addressed before merge.

shyjsarah:
Re-reviewed the latest head (`314c6bf`). The implementation direction is good, but I found three blockers, so I don't think this is ready to merge yet:

1. **Add-time encoding and search-time scoring use inconsistent L2 arithmetic.** The new blocked encoder minimizes direct squared differences, while IVF-PQ search still uses the cancellation-prone `||q||² + ||c||² - 2q·c` table. I reproduced this through the public train/add/write/open/search APIs using finite data: an exact indexed query row was scored at `8388608`, while a farther row was scored at `0` and returned first. The merge base did not show this reversal.

2. **The `<32` direct path regresses x86_64 small-batch adds.** For the included 31-row `d=768,m=192` shape, an independent optimized x86_64 run measured about `3.63 ms` on this PR versus `1.11 ms` at the merge base (>3x slower). The generic `f32::mul_add` loop compiles to inner-loop `fmaf` calls on x86_64.

3. **`dsub>4` also regresses on x86_64.** The transposed path is enabled for every `dsub>=4`, but SIMD specialization only handles `dsub==4`; the documented 32,768-row `dsub=8` case measured about `212 ms` on this PR versus `74 ms` at the merge base (~3x slower).

Suggested direction:

- Use one numerically consistent distance definition for both PQ assignment and all IVF-PQ search-table paths, with an end-to-end large-offset regression test.
- Keep the previous x86 path for small batches and unsupported `dsub` shapes until suitable AVX2 kernels are available, or introduce target-specific thresholds/kernels.
- Follow-up: add required aarch64 core-test coverage for the new unsafe NEON implementation.

Local verification otherwise passed: core tests (465 passed, 1 ignored), targeted PQ tests, workspace Clippy, fmt check, and benchmark compilation.

core/src/ivfpq.rs
```diff
@@ -252,7 +252,7 @@ impl IVFPQIndex {
 
         let code_size = self.pq.code_size();
         let mut codes = vec![0u8; n * code_size];
-        self.pq.encode_batch(&to_encode, n, &mut codes);
+        self.pq.encode_batch_blocked(&to_encode, n, &mut codes);
```
shyjsarah: **[Major correctness]** This switches persisted code assignment to direct squared differences, but IVF-PQ search still scores those codes with the norm/dot-product distance table. Those arithmetic definitions can disagree materially, not just on ulp-level ties.

I reproduced an end-to-end public API case with finite values where exact query rows are assigned code 1, but search scores code 1 as `8388608` and a farther code-0 row as `0`, returning the farther row first. The merge base returns an exact row first.

Please make assignment and all IVF-PQ search-table paths use the same numerical distance definition, and add the train/add/write/open/search reproduction as a regression test.

core/src/pq.rs
```diff
@@ -370,6 +376,113 @@ impl ProductQuantizer {
...
+    /// from coordinate differences, avoiding the cancellation possible in the
+    /// norm/dot identity used by [`Self::encode_batch`]. The different arithmetic
+    /// can produce different valid codes, so use this only where codes are freshly
+    /// produced (index build), not where byte-stable output is pinned. Trained
+    /// codebooks are expected to contain only finite values.
+    pub(crate) fn encode_batch_blocked(&self, data: &[f32], n: usize, codes: &mut [u8]) {
+        // The 4-bit packed path keeps the original per-vector implementation.
+        if self.nbits == 8 && (0..self.m).all(|sub| self.chunk_dim(sub) >= 4) {
+            if n < ENCODE_TRANSPOSE_MIN_ROWS {
+                self.encode_batch_8bit_direct(data, n, codes);
```
shyjsarah: **[Major x86_64 performance regression]** This unconditional `<32` branch routes small adds through the generic per-centroid direct loop. On optimized x86_64, its inner `f32::mul_add` lowers to repeated `fmaf` calls.

For the included 31-row `d=768,m=192` shape, I measured about `3.63 ms` here versus `1.11 ms` at the merge base (>3x slower). Please retain the previous batch path on x86 unless a target-specific direct kernel is available, or choose the threshold per backend.

core/src/pq.rs
```diff
@@ -527,6 +640,189 @@ fn argmin_code(distances: &[f32]) -> u8 {
...
+        if q.len() == 4
+            && ksub.is_multiple_of(8)
+            && is_x86_feature_detected!("avx2")
+            && is_x86_feature_detected!("fma")
+        {
+            // SAFETY: AVX2 and FMA presence checked above; slice bounds checked by caller.
+            return unsafe { score_argmin_avx2_d4(q, t, ksub) };
+        }
+    }
+    score_argmin_scalar(q, t, ksub, scores)
```
shyjsarah: **[Major x86_64 performance regression]** The outer gate enables this transposed path for every `dsub>=4`, but only `dsub==4` has an AVX2/NEON specialization. `dsub=8` reaches this scalar fallback, whose `mul_add` also becomes inner-loop `fmaf` calls on optimized x86_64.

For the documented 32,768-row `dsub=8` case, I measured about `212 ms` here versus `74 ms` at the merge base (~3x slower). Please keep unsupported shapes on the matrixmultiply path until a wider SIMD kernel is available, or add an arbitrary-`dsub` AVX2 implementation.

shyjsarah:
Re-reviewed the latest head (`7296eaa`) after the eight follow-up commits since `314c6bf`.

The generic AVX2/FMA work fixes the previously reported `dsub>4` regression on capable x86_64 hosts, and the new per-subspace exact fallback improves the large-offset/precomputed-table behavior. The new targeted tests, core suite, and workspace Clippy pass locally.

Two blockers remain:

1. **Add/search numerical consistency is still not fixed across all paths.** The direct/FMA 8-bit encoder can choose code 1 while the ordinary, unoptimized-reader, ephemeral-reuse, resident-precomputed, and optimized serialized-reader search paths all rank code 0 first. The new fallback only detects individual distances near the cancellation bound; it does not protect a small pairwise margin between two large nonzero distances. I reproduced the same failure on `7296eaa` through all five paths.

2. **Small adds still regress on supported x86_64 hosts without AVX2/FMA.** `encode_batch_blocked` now always transposes the full codebook. For `d=768,m=192`, this writes about 0.75 MiB before encoding any row. On the scalar runtime fallback, rows 1/7/8/31 were approximately 39%/46%/37%/45% slower than `314c6bf`; the regression disappears at 32 rows, isolating the fixed transpose cost.

Minor follow-up: the public distance-table fallback now uses direct/FMA arithmetic in cancellation cases, while 4-bit and byte-stable encoders still select codes with the legacy norm/dot arithmetic. An extreme finite fixture changes top-1 after upgrading the scorer, including after write/reopen. I suggest separating the legacy/byte-stable table contract from the direct-encoded 8-bit IVF-PQ table contract, or versioning the scoring semantics.

I still recommend holding the merge until the two blockers above are addressed. Cross-architecture parity coverage for NEON, generic AVX2/FMA, and the scalar fallback would also be valuable.

core/src/pq.rs
```diff
@@ -436,7 +535,29 @@ impl ProductQuantizer {
                             let c_off = c_base + j * chunk_dim;
                             fvec_norm_l2sqr(&self.centroids[c_off..c_off + chunk_dim])
                         };
-                        table[t_base + j] = (q_norm + c_norm - 2.0 * table[t_base + j]).max(0.0);
+                        let approximate = q_norm + c_norm - 2.0 * table[t_base + j];
+                        // Cover the three reductions and final subtraction before trusting SGEMM.
+                        let error_bound =
+                            16.0 * chunk_dim as f32 * f32::EPSILON * (q_norm + c_norm);
+                        table[t_base + j] =
+                            if !approximate.is_finite() || approximate <= error_bound {
```
shyjsarah: **[Major correctness]** This fallback only asks whether each individual expanded distance is near its own cancellation bound. It does not protect the ordering when two large, nonzero candidate distances differ by less than their numerical error.

I reproduced this on `7296eaa`: the direct/FMA add path selects code 1 (`88314130 < 88314136`), but ordinary in-memory search, unoptimized reader search, ephemeral reuse, resident precomputation, and optimized serialized-reader search all rank code 0 first. The new per-sub exact fallback is not triggered for this subspace.

Please make the direct-encoded 8-bit table contract canonical across all search paths, or use a pairwise error/margin rule that recomputes candidates whose ordering is not provably stable. The ephemeral f64-to-f32 path also needs to preserve the same f32/FMA ordering used for persisted code assignment.

core/src/pq.rs
```diff
@@ -370,6 +386,87 @@ impl ProductQuantizer {
...
+    /// their distance-table memory traffic. Distances are accumulated directly
+    /// from coordinate differences, avoiding the cancellation possible in the
+    /// norm/dot identity used by [`Self::encode_batch`]. The different arithmetic
+    /// can produce different valid codes, so use this only where codes are freshly
+    /// produced (index build), not where byte-stable output is pinned. Trained
+    /// codebooks are expected to contain only finite values.
+    pub(crate) fn encode_batch_blocked(&self, data: &[f32], n: usize, codes: &mut [u8]) {
+        // The 4-bit packed path keeps the original per-vector implementation.
+        if self.nbits == 8 && (0..self.m).all(|sub| self.chunk_dim(sub) >= 4) {
+            self.encode_batch_8bit_transposed(data, n, codes);
```
shyjsarah: **[Major x86_64 fallback regression]** This now transposes the full codebook for every eligible add, even when runtime dispatch selects the scalar scorer. For `d=768,m=192`, that is about 0.75 MiB of allocation/writes before processing a single row.

On a supported x86_64 runtime without AVX2/FMA, versus `314c6bf`, I measured approximately +39%/+46%/+37%/+45% for rows 1/7/8/31; rows=32 was effectively unchanged, isolating the fixed transpose cost.

Please restore a small-batch direct path when `score_argmin_kernel` resolves to scalar, use a backend-specific threshold, or cache the transposed codebook after training.

shyjsarah:
Re-reviewed the latest head (`9e6cb1b`). CI is green, the core suite passes locally (471 passed / 1 ignored), and the previous no-AVX2/FMA small-add transpose regression is fixed by restoring the direct path.

The new pairwise-margin certificate fixes the earlier two-code fixture, but I found a remaining correctness hole plus two search-performance blockers:

1. **The certificate proves only the global table argmin, while IVF-PQ scans arbitrary persisted codes.** Adding an unused third centroid that is a stable global argmin suppresses exact fallback, while stored codes 0 and 1 remain reversed. On the latest head, add emits `[0,1]` and direct/FMA scoring ranks id 1 first, but ordinary in-memory, reopened reader, forced ephemeral reuse, resident optimized reader, and in-memory precomputed search all return id 0.

2. **The certificate broadly regresses standard-scale L2 search.** Compared with `7296eaa`, a short standard-scale benchmark measured ordinary search about +36%, resident precomputed about +56%, and ephemeral reuse about +217%. Each completed 256-entry sub-table is rescanned by `min_by` and `all()`, with repeated error-bound work in the hot table-building loops.

3. **Unstable ephemeral fallback pays both algorithms and allocates per query/list.** Large-offset Auto ephemeral search measured about `6.6 ms` versus `0.46 ms` at `7296eaa` (~14x). The fallback makes exactly `nq*nprobe` extra residual-vector allocations and recomputes the complete exact table after ephemeral list/query tables were already built.

Suggested direction:

- Do not use stability of the global centroid argmin as a proxy for the ordering of arbitrary stored codes. Either use the canonical direct/FMA table or certify the comparisons actually consumed by scanning.
- Fuse stability tracking into table construction instead of rescanning completed tables.
- Reuse a per-worker residual buffer and avoid selecting ephemeral reuse when the scale predicts a high exact-fallback rate.

Minor follow-up: the small-batch dispatch policy would be more robust with an explicit kernel-kind enum rather than function-address equality.

I still recommend holding the merge until the three blockers above are addressed.

jerry-024:
Rewrote the branch to remove the root cause of the repeated stability blockers instead of patching them.

The transposed-codebook kernel minimized direct squared differences while search scored codes via the `||q||² + ||c||² − 2q·c` expansion. Two different roundings of the same L2 distance meant add and search could disagree on argmin ties, and each stability certificate / exact fallback only covered the cases it was built for — @shyjsarah kept finding valid new ones because the inconsistency was structural.

The new encoder computes distances with one SGEMM per sub-quantizer block using exactly the same identity as `encode_batch` and the search-time tables, so codes are byte-identical to the canonical per-vector path and the add/search inconsistency cannot occur. All the certificate/fallback machinery (`fvec_madd_two_minima`, per-sub exact fallback, serialized norm bounds) is deleted; the diff is now +386/−1.

Fresh 8-thread benchmarks on the same M3 Pro (baseline = merge-base `b5b4a77`): 1.69x at 4096 rows (`d=768, m=192`), 2.16x at 32768 rows (`dsub=8`, which SGEMM handles natively rather than via scalar fallback), no small-batch regression below the 32-row threshold. Updated numbers are in the PR description.

core/src/pq.rs
```diff
@@ -370,6 +376,85 @@ impl ProductQuantizer {
         );
     }
 
+    /// Blocked batch encode for the IVF-PQ add path.
+    pub(crate) fn encode_batch_blocked(&self, data: &[f32], n: usize, codes: &mut [u8]) {
```
shyjsarah, line 380: **Suggestion (non-blocking):** The fixed `n >= 32` SGEMM threshold may be too low for wide Rayon pools, where small batches are split into tiny SGEMM calls. In a local 32-thread benchmark, 32/64-row batches were slower than the per-vector path. Could the threshold account for the pool width, or could we add benchmarks covering small batches with different thread counts?

shyjsarah:
+1

PR #85, opened 2026-08-28
perf: accelerate IVF builds with Vamana coarse assignment

jerry-024:
## Summary

Speed up large IVF index builds by assigning vectors to IVF lists through a small reusable Vamana graph. The optimized coarse-assignment path covers IVF-Flat, IVF-PQ, IVF-SQ, and IVF-RQ. Exact assignment remains in place for smaller centroid sets and as a fallback when graph construction fails.

## Changes

- Add shared coarse-assignment state for every IVF builder:
  - IVF-Flat `add`
  - IVF-PQ residual training and `add`
  - IVF-SQ residual training and `add`
  - IVF-RQ `add`
- Build a Vamana graph when `dimension * nlist >= 1,000,000` and reuse it across batches.
- Keep IVF centroid storage private and expose validated getters/setters so centroid replacement always invalidates the cached graph; IVF-RQ also rebuilds centroid norms and rotated centroids.
- Avoid a second resident centroid copy and the full centroid comparison on every assignment.
- Reuse bounded Vamana search scratch state across assignments and cap it to the graph size.
- Keep batched coarse top-k on SGEMM while:
  - tiling queries so the SGEMM result matrix stays within 4M `f32` values (about 16 MiB),
  - directly refining candidates within the floating-point error band around the `nprobe` boundary,
  - returning direct L2 distances for selected centroids,
  - falling back to complete direct comparison for non-finite norms, products, or expanded distances,
  - computing the error scale with one square root per query instead of one per query-centroid pair.
- Use a degree-clamped sequential Vamana memory estimate without parallel-only buffers.
- Parallelize cosine normalization for batches larger than 1,000 vectors.
- Reuse per-worker RQ encoding scratch buffers.
- Add opt-in IVF-RQ train, add, and serialization stage timings through `PAIMON_LOG_VECTOR_INDEX_BUILD_TIMING`.

## Benchmark

Single-run results on an Intel Xeon 6982P-C machine with 8 cores, 16 threads, and 61 GiB of RAM, using 10M 768-dimensional cosine vectors, `nlist=4096`, direct OSS reads, and one output index shard (`rq_bits=4` for IVF-RQ and `m=192` for IVF-PQ). The table reports Paimon's `index_add_ms` stage, which is directly affected by this change; it includes coarse assignment and the remaining index-specific add/encoding work, not end-to-end build time:

| Index | Exact assignment | Vamana assignment baseline | Change |
| --- | ---: | ---: | ---: |
| IVF-Flat | 110.887 s | 60.730 s | -45.2% |
| IVF-PQ | 316.324 s | 262.635 s | -17.0% |
| IVF-SQ | 91.939 s | 41.374 s | -55.0% |
| IVF-RQ | 117.508 s | 67.008 s | -43.0% |

An isolated release-mode coarse top-k microbenchmark on an Apple M3 Pro (11 logical CPUs) used five runs and reports median time:

| Shape (`nq × nlist × d`) | Before tiling/refinement (`38c22db`) | Current | Change |
| --- | ---: | ---: | ---: |
| `256 × 4096 × 768` | 60.61 ms | 18.99 ms | -68.7% |
| `8192 × 4096 × 16` | 187.80 ms | 113.27 ms | -39.7% |

For the second shape, peak RSS fell from 134.1 MiB to 22.3 MiB (-83.4%). The current single-pass non-finite check remains in the same timing range as `c985646` (18.76 ms and 116.52 ms respectively), while preserving complete direct fallback. This microbenchmark isolates coarse top-k and does not replace the 10M end-to-end query run.

Post-change query A/B reused the same 10M IVF-RQ Vamana index with direct OSS reads. The baseline binary used `06ade13`; the current binary used `0336679`. Each point is the pooled median of two process runs with three measured trials each: 50 queries, 5 warm-up queries, `top_k=100`, individual search, request concurrency 8, shard concurrency 32, and 64 range-read threads.

| nprobe | Recall@100 | Baseline QPS | Current QPS | Change |
| ---: | ---: | ---: | ---: | ---: |
| 32 | 0.8048 | 25.60 | 25.01 | -2.3% |
| 128 | 0.9210 | 11.02 | 10.95 | -0.6% |
| 256 | 0.9474 | 6.29 | 6.19 | -1.5% |
| 1024 | 0.9598 | 1.63 | 1.64 | +0.4% |

## Testing

- [x] `cargo test -p paimon-vindex-core` (467 unit tests and 9 integration tests passed; 2 ignored)
- [x] `cargo clippy -p paimon-vindex-core --all-targets -- -D warnings`
- [x] `cargo build --workspace`
- [x] Regression coverage for ambiguous `nprobe` boundaries, translated and non-finite batch top-k, tiled SGEMM scratch, and sequential Vamana memory budgeting
- [x] Release-mode coarse top-k latency and peak-RSS microbenchmark
- [x] 10M IVF build-stage benchmarks comparing exact and Vamana assignment across IVF-Flat, IVF-PQ, IVF-SQ, and IVF-RQ
- [x] IVF-RQ post-change 10M query A/B across nprobe 32, 128, 256, and 1024

## Notes

- Approximate assignment is only enabled for large centroid matrices; smaller indexes keep exact assignment.
- IVF builder centroid fields are now private in favor of getters/setters. This is a source-level API change for 0.5.0: replace direct reads with `quantizer_centroids()` and assignments with `set_quantizer_centroids(...)`. The stored index format is unchanged.
- The 10M build-stage figures are single-run measurements. Query A/B used two process replicates with three trials each and was measured only on IVF-RQ.
- The post-change 10M query A/B preserved Recall@100 exactly; pooled median QPS changed by -2.3% to +0.4% across the measured nprobe values.
- Relates to #82.


core/src/coarse.rs
```diff
@@ -0,0 +1,154 @@
...
+
+        let mut assignments = vec![0usize; n];
+        let chunk = (n / (rayon::current_num_threads() * 4).max(1)).clamp(16, 1024);
+        assignments.par_chunks_mut(chunk).enumerate().for_each_init(
+            || graph.search_scratch(APPROX_ASSIGN_SEARCH_LIST),
+            |scratch, (chunk_idx, chunk_assignments)| {
+                let row0 = chunk_idx * chunk;
+                for (i, assignment) in chunk_assignments.iter_mut().enumerate() {
+                    let row = row0 + i;
+                    *assignment = graph
```
shyjsarah, line 121: **[major] The list chosen here is approximate, but all IVF search paths still choose probe lists with exact centroid top-k. The two policies can disagree, so a vector may be stored in a list that a self-query with `nprobe=1` never probes. I reproduced this at `d=256, nlist=4096`: 1,807/4,096 centroid vectors moved away from their exact list, and id 2461 was stored in list 3 while its self-query returned `-1 / f32::MAX`. Please keep one coarse-quantizer contract for both insertion and lookup, or make this approximation explicit and ensure search covers the graph-selected list. A threshold-enabled self-recall regression test would catch this.

shyjsarah (changes requested):
Thanks for the detailed benchmarks and regression coverage. I found one correctness blocker and two smaller regressions:

1. **Blocker:** above the automatic threshold, add/train persists the Vamana-selected coarse list while search still probes exact centroid top-k lists. This makes the assignment and lookup contracts inconsistent and can make an inserted vector unreachable. In a release-mode reproduction with `d=256, nlist=4096`, 1,807/4,096 centroid vectors were stored outside their exact list; a self-query for id 2461 with `k=1, nprobe=1` returned `-1 / f32::MAX`.
2. **Minor performance regression:** the first assignment builds the whole graph without considering `n`; at `d=512, nlist=2048`, `n=1` took about 30.8 ms versus 0.133 ms for exact assignment (~232x). Empty batches also pay this setup cost.
3. **Minor boundary regression:** `preprocess_vectors(..., d=0, Cosine)` now panics in both `chunks_mut(0)` and `par_chunks_mut(0)`, whereas the previous indexed loop returned an empty vector.

Please make add and search use a consistent coarse-assignment policy (keeping exact assignment as the default or making graph assignment explicit would be the safest option) and add threshold-enabled self-recall coverage. The two minor issues can be handled with an empty/small-batch policy and a `d == 0` guard.

jerry-024:
Thanks for the detailed review. Addressed in `8c706af`.

1. We are keeping automatic Vamana coarse assignment as an explicit build-speed/recall tradeoff rather than changing search-side list selection. All option-map APIs now accept `ivf.coarse-assignment=auto|exact`. `auto` preserves the thresholded Vamana behavior; `exact` disables the graph and restores exact nearest-centroid assignment. Rust config literals expose the same choice through `use_approximate_coarse_assignment`, and the four direct IVF index types expose `set_approximate_coarse_assignment(...)`. The API and per-index docs now state that automatic graph assignment may select a different list from exact search, can reduce recall at small `nprobe`, and does not guarantee self-recall at `nprobe=1`. A threshold-enabled regression test verifies that `exact` preserves self-recall with `d=256`, `nlist=4096`, and `nprobe=1`.
2. Empty assignments now return before graph preparation, including the separate IVF-RQ preparation path. Small non-empty batches still pay the graph startup cost when automatic assignment is enabled; this is documented, and users can select `exact` to avoid it.
3. `preprocess_vectors` now skips cosine chunking when `d == 0`, with coverage for both the serial and parallel thresholds.

Verification: `cargo test --workspace` passed 496 tests with 2 ignored; `cargo clippy --workspace --all-targets -- -D warnings`, formatting, and `git diff --check` all passed.

core/src/index.rs
```diff
@@ -600,6 +625,20 @@ fn parse_nlist_options(
     }
 }
 
+fn parse_ivf_coarse_assignment_option(options: &mut ConfigOptions) -> io::Result<bool> {
+    match options
+        .optional("ivf.coarse-assignment")
+        .as_deref()
+        .map(str::trim)
+    {
+        None | Some("auto") => Ok(true),
```
shyjsarah, line 634: **[major] This leaves the original blocker on the default path.** The new `exact` option works, but omitted/`auto` still assigns through Vamana while search selects lists using exact centroid top-k. Re-running the original release probe on this commit still moved 1,807/4,096 centroid vectors and returned `-1 / f32::MAX` for a self-query with `nprobe=1`; explicit exact mode moved 0 and succeeded. Documenting the loss makes it visible, but does not make add and lookup use a consistent default contract. Please default to exact or unify the query-side policy.

core/src/ivfpq.rs
```diff
@@ -114,9 +116,40 @@ impl IVFPQIndex {
...
     }
 
+    pub fn quantizer_centroids(&self) -> &[f32] {
+        &self.quantizer_centroids
+    }
+
+    /// Enables automatic Vamana coarse assignment for large centroid matrices.
+    /// Disable it to keep vector assignment exact.
+    pub fn set_approximate_coarse_assignment(&mut self, enabled: bool) {
+        assert!(
```
shyjsarah, line 130: **[minor] This guard permits changing the policy after `train()` (or on an index created by `from_trained`) as long as no rows were added.** For residual IVF-PQ, training uses the current coarse assignments to build the PQ training residuals, so switching here makes add encode a different assignment distribution than the one used for training. IVF-SQ has the same issue for its residual/list quantizers. Please restrict policy changes to pre-training state, or invalidate/retrain assignment-dependent state; also document the lifecycle requirement.

shyjsarah (changes requested):
Thanks for adding an explicit coarse-assignment policy and the regression coverage. I re-ran the original reproductions on `8c706af`:

- **Original correctness blocker: not fixed on the default path.** `ivf.coarse-assignment=exact` is an effective opt-out, but omitted/`auto` still uses approximate assignment for add while search probes exact centroid top-k. With the same `d=256, nlist=4096` release reproduction, default mode still moved 1,807/4,096 centroid vectors and the self-query returned `-1 / f32::MAX`; exact mode moved 0 and returned the vector. The docs now disclose the behavior, but the default add/search contract remains inconsistent.
- **Small-batch performance: partially fixed.** `n=0` now returns before graph construction. Default `n=1` still builds the graph: about 30.4 ms versus 0.053 ms for direct exact assignment in my rerun. Explicit exact mode took about 0.073 ms.
- **Zero-dimensional cosine panic: fixed.** Both serial and parallel reproductions now complete without panic.

I also found one lifecycle issue in the new low-level setter: IVF-PQ/IVF-SQ can change the assignment policy after training but before the first add, although their residual/codebook state was trained using the previous policy. Please either restrict this setter to pre-training state or retrain/invalidate the assignment-dependent model state.

Local validation: core tests passed (`472 passed, 1 ignored`; integration `9 passed, 1 ignored`) and clippy passed.

core/src/ivfpq.rs
jerry-024: Fixed in `0dd8915`. The coarse-assignment setters on all four direct IVF index types are now crate-private, so callers cannot switch policy after training (or on an index returned by `from_trained`). Public callers select the policy through `VectorIndexConfig` or `ivf.coarse-assignment=auto|exact` before the writer/training lifecycle starts. `from_trained` only copies the source index policy internally. The API and release documentation now state this lifecycle.

core/src/index.rs
jerry-024: I investigated the corresponding Lance behavior and am intentionally retaining `auto` as the default because it aligns with Lance’s existing IVF build contract introduced by [lance-format/lance#4089](https://github.com/lance-format/lance/pull/4089):

- Lance defaults its indexing speedup status to `Auto` and enables it when `nlist × dimension >= 1,000,000` ([source](https://github.com/lance-format/lance/blob/742e6a317c95840ca18adf47646ea79e0070c57b/rust/lance-index/src/vector/utils.rs#L24-L103)).
- Build-time partition assignment uses HNSW top-1 ([source](https://github.com/lance-format/lance/blob/742e6a317c95840ca18adf47646ea79e0070c57b/rust/lance-index/src/vector/ivf/transform.rs#L46-L116)), while query-time `find_partitions` still uses exact centroid top-k ([source](https://github.com/lance-format/lance/blob/742e6a317c95840ca18adf47646ea79e0070c57b/rust/lance-index/src/vector/ivf.rs#L343-L353)).
- Lance provides an exact opt-out through `LANCE_USE_HNSW_SPEEDUP_INDEXING=disabled`.

Paimon follows the same policy boundary and threshold, using Vamana (`R=12`, `L=15`) instead of HNSW (`M=12`, `ef=15`), and provides the finer-grained `ivf.coarse-assignment=exact` opt-out. The low-`nprobe` effect is therefore an intentional, documented build-speed/recall tradeoff of `auto`; callers requiring exact nearest-centroid placement can select `exact`.

shyjsarah:
+1

PR #87, opened 2026-09-02
feat: allow training results to create multiple writers

### LINKED ISSUE #86: Support reusing a training result across multiple vector index writers
## Motivation

Callers that build one vector index file per segment currently need a separate training pass for every writer. The trained centroids, quantizers, and rotations can often be shared when those segments come from the same table or data distribution, but the unified API consumes `VectorIndexTraining` when it creates a writer.

For `N` segments, this makes the lifecycle effectively:

```text
train -> writer 1
train -> writer 2
...
train -> writer N
```

IVF-PQ already has an internal `from_trained` path, but this capability is not available through the unified API or consistently across the supported index types and language bindings.

## Proposed API

Allow a completed training result to create multiple independent writers:

```rust
let training = VectorIndexTrainer::train(config, training_vectors, training_count)?;
let first = training.create_writer();
let second = training.create_writer();
```

Each writer should start with the same trained model and an empty vector/ID payload. The existing consuming APIs should remain unchanged for one-shot callers.

Equivalent opt-in entry points should be exposed through C, C++, Java/JNI, and Python.

## Expected benefit

In a release-mode synthetic build with 128-dimensional vectors, 8,192 training rows, and eight segments of 2,048 vectors each, training once reduced writer preparation time by 84-87% across IVF-Flat, IVF-SQ, IVF-PQ, IVF-RQ, and DiskANN. Including vector insertion and serialization, elapsed time fell by 73-84%.

The benefit comes from reducing `N` training passes to one; it does not change the index algorithms themselves.

## Compatibility and scope

- No default behavior change: callers opt in to reusable training explicitly.
- Existing consuming APIs remain available.
- No on-disk format change.
- No cross-process training-model serialization or index merging in this issue.

wangzhigang1999:
## Summary

Allow one completed `VectorIndexTraining` to create multiple independent writers. Callers building several segment or file indexes from the same data distribution can train once and reuse the trained centroids, quantizers, and rotations instead of repeating training for every writer.

The existing consuming APIs remain unchanged for one-shot callers. Reusable writer creation is explicit and opt-in.

Closes #86.

## Changes

- Add `VectorIndexTraining::create_writer()` for IVF-Flat, IVF-SQ, IVF-PQ, IVF-RQ, and DiskANN.
- Initialize each reusable writer with the trained model and empty segment-local IDs, vectors, codes, and graph state.
- Expose equivalent non-consuming entry points through C, C++, Java/JNI, and Python while retaining the existing consuming forms.
- Document both lifecycle choices without changing the primary one-shot examples.
- Cover independent writer payloads with write/open/search round trips and keep binding tests focused on their ownership and ABI boundaries.

## Benchmark

Release-mode synthetic benchmark on an Intel Xeon Platinum 8575C host with 32 logical CPUs and 60 GiB RAM. The workload used 128-dimensional vectors, 64 IVF lists, 8,192 training rows, eight segments, and 2,048 vectors per segment. `retrain` trains separately for every segment; `reuse` trains once and creates eight writers. The table reports the median of three alternating process runs.

| Index | Prepare: retrain -> reuse | Speedup | Full build: retrain -> reuse | Speedup |
| IVF-Flat | 112 -> 18 ms | 6.2x | 138 -> 33 ms | 4.2x |
| IVF-SQ | 123 -> 20 ms | 6.2x | 140 -> 27 ms | 5.2x |
| IVF-PQ | 1,355 -> 173 ms | 7.8x | 1,384 -> 194 ms | 7.1x |
| IVF-RQ | 109 -> 16 ms | 6.8x | 132 -> 28 ms | 4.7x |
| DiskANN | 1,240 -> 161 ms | 7.7x | 1,485 -> 407 ms | 3.6x |

The preparation measurement isolates training and writer creation. The full-build measurement also adds and serializes all segment vectors. Serialized output sizes were unchanged for each index type. This benchmark measures the removal of repeated training work; it does not claim an index-algorithm speedup.

## Testing

- [x] `cargo test --workspace --locked` (487 passed, 2 ignored)
- [x] `cargo clippy --all-targets --workspace --locked -- -D warnings -A clippy::chunks-exact-to-as-chunks`
- [x] C tests (10 passed)
- [x] C++ tests (8 passed)
- [x] Java Maven tests and JNI native validation
- [x] Python 3.12 tests (29 passed)

The Clippy allow is limited to the `chunks_exact_to_as_chunks` lint newly enabled by Rust 1.98; strict Clippy currently reports it at 28 unchanged locations on `main`. No call site introduced by this PR requires the allow.

## Compatibility

- No default behavior change; reuse requires an explicit new API call.
- Existing consuming APIs remain source compatible.
- No on-disk format change.
- No cross-process training-model serialization or index merging.


python/paimon_vindex/__init__.py
```diff
@@ -512,6 +521,15 @@ def __init__(self, training: VectorIndexTraining):
             _check_error("failed to open writer")
         self._dimension = self._read_dimension()
 
+    @classmethod
+    def _from_handle(cls, handle):
+        writer = cls.__new__(cls)
```
JingsongLi: [P2] Free the native handle if wrapper allocation fails

At this point `create_writer()` already owns a newly allocated native writer, but `cls.__new__()` and `_NativeHandleLock()` can raise before `writer._handle` is installed. Under memory pressure, no object owns the raw pointer, so `__del__` cannot recover it and the cloned model is leaked. Please make this method an explicit ownership-transfer boundary: catch `BaseException` and call `paimon_vindex_writer_free(handle)` exactly once when construction fails. Fault-injection coverage should exercise failures both before and after `_handle` assignment. The existing `_read_dimension()` comment covers a later failure, but this earlier window is the unrecoverable one.

wangzhigang1999: Thanks @JingsongLi, fixed in ddf6b4a. I added fault-injection tests before and after handle assignment, and all CI checks pass. Could you take another look when you have time?

Issue #97, opened 2026-09-08
[Feature] Distance range search for IVF indexes


JunRuiLee:
## Motivation

The unified reader exposes only top-K search. A predicate such as
`vector_distance(col, q) < 0.5` therefore has to be answered by asking for a
large K and filtering afterwards, which either over-reads (K too large) or
silently drops matching rows (K too small). Neither is acceptable when the
predicate is pushed down from SQL, because the result set no longer equals the
predicate.

This is not a new idea for the families this crate mirrors. FAISS declares
`range_search` on `Index` itself and implements it across the IVF families, the
flat-codes indexes and HNSW, so the general semantics -- what a hit set
contains, how the boundary is treated, what the result container looks like --
have an established reference. The per-family internals differ enough from this
crate's that they are a reference for semantics, not for implementation. Where
this proposal deviates from FAISS it is called out explicitly (the two-sided
band in question 1 below, and the out-of-scope list at the end).

## Scope of the proposal

Add distance-range search to the unified reader, alongside the existing top-K
entry points and following their existing shape (single and batch queries, with
and without a serialized Roaring filter).

Intended coverage, and where each family's reference behaviour comes from:

| Family | Range membership decided on | Closest FAISS analogue for semantics |
|---|---|---|
| IVF-Flat | exact distances | `IndexIVFFlat` |
| IVF-SQ | quantized estimate | `IndexIVFScalarQuantizer` |
| IVF-PQ | quantized estimate | `IndexIVFPQ` |
| IVF-RQ | quantized estimate | `IndexIVFResidualQuantizer` (additive-quantizer family) |
| DiskANN | **not proposed** -- see out of scope | -- |

Metrics: L2 first, then cosine and inner product (a later step). Both single and
batch queries, with and without a Roaring filter, for every supported family --
no entry point is left as a stub that silently behaves differently from its
top-K counterpart.

Two decisions shape the API rather than the implementation, so I would rather
settle them here than in code review. I have a leaning on both and give the
reasoning, but neither is something I am attached to:

1. **Should the index layer take a distance band, or SQL operators?** A band
   with two optional bounds keeps engine-specific operator vocabulary out of the
   index layer, at the cost of asking callers to derive the boundary value.
   Knowhere is the precedent for two bounds (`radius` + `range_filter`); FAISS,
   Doris and OpenSearch each expose a single threshold instead.
2. **How should the estimator families report incompleteness?** IVF-Flat can
[...]

PR #99, opened 2026-09-10
feat: distance range search for IVF-Flat

JunRuiLee:
Adds range search to the unified reader alongside top-K: given a distance
band, return every row of the probed lists whose distance falls inside it.
There is no K and no truncation, aligned with Faiss's range_search. Results are
neither sorted nor padded, because ordering is the caller's business.

Scope is IVF-Flat under L2. Every other family reports a distinguishable
error rather than silently behaving differently from its top-K counterpart, so
a caller can fall back to an exhaustive scan. Top-K plus post-filtering is not
an equivalent fallback and is not offered as one: it reintroduces the very
truncation this feature removes, since a band matching more rows than K cannot
return them. All four entry point shapes are covered -- single and batch, with
and without a serialized Roaring filter -- so none is left as a stub.

"No cap" is a promise about not truncating, not about completeness. IVF-Flat
computes exact distances for the rows it visits, but with nprobe < nlist a
matching row in an unprobed list is still missed. Removing K removes
truncation; it does not make the search exhaustive.

The scan seam
-------------

Top-K and range want the same per-list scan and differ only in what they do
with a row, so scan_flat_list and scan_flat_rows become generic over a
Collector trait: it reports an admission threshold, receives kept rows, and
is told when a row is abandoned early. ReaderTopKHeap and RangeCollector are
its two implementations, aligned with Faiss's InvertedListScanner, which hands
rows to a ResultHandler& in scan_codes. Faiss splits the range case out into a
separately named scan_codes_range taking a RangeQueryResult&; one collector
serves both here, so there is a single kernel rather than a pair to keep in
step.

The seam is IVF-Flat's only. The other families have their own scan kernels
and are untouched here; each will generalize its own when its range support
lands, against this same trait.

Two things about top-K are unchanged, and one changes:

- ReaderTopKHeap::cutoff() is worst_distance().unwrap_or(INFINITY). An unfilled
  heap yields an infinite cutoff, which nothing can exceed, so such a row is
  computed rather than abandoned -- as before, when the abandon path sat inside
  `if let Some(threshold) = heap.worst_distance()`.
- note_abandoned() has an empty default body and ReaderTopKHeap does not
  implement it. It is on the trait because an abandoned row never reaches
  push(), so a collector counting scanned rows cannot recover the count
  unless the kernel reports it.
- Collector::push is fallible, and the streamed-chunk callback used by the
  oversized-list path becomes fallible with it, so a collector error can
  propagate out instead of having to panic.

What changes is that top-K now prunes exactly, which is a fix. Its cutoff is a
distance the heap computed with fvec_l2sqr, while the old pruning kernel
accumulated differently; a candidate could therefore be abandoned for passing a
cutoff its committed distance would not have passed. The range path covered that
divergence with a widening term, but on the top-K path there was no widening at
all. Pruning against the accumulation that will be committed removes the
divergence rather than budgeting for it, so recall can only improve. Returned
labels and distances are unchanged -- a committed distance is still exactly
fvec_l2sqr's -- and the existing top-K tests pass unmodified.

The defect lives in the kernel, so that is where its guard is:
distance::tests::a_distance_equal_to_the_threshold_is_never_abandoned puts the
cut on each row's own distance, the tightest admissible case, and requires every
row back with identical bits.

Range semantics
---------------

The API takes a two-sided band (DistanceBand with two optional Bounds)
rather than SQL operators, keeping engine-specific operator vocabulary out
of the index layer; DistanceEndpoint plus CutOperator derive the boundary
for callers that think in operators. Knowhere is the precedent for two
bounds.

Membership is decided by the band on the committed distance, and the abandon
threshold is the raw upper cut with no margin.

That is sound because pruning and committing are one accumulation.
fvec_l2sqr_unless_exceeds returns the squared L2 distance unless its running sum
passes the threshold first, and the value it returns is bit-identical to
fvec_l2sqr's: it is the same accumulator in the same order with the same
reduction, read without disturbing it every 128 elements. Adding a non-negative
f32 is non-decreasing under IEEE rounding and the reduction is monotone in every
lane, so a partial sum above the cut proves the completed distance is above it.

Doing it with two kernels was tried first and is what the earlier revisions of
this branch were about. A pruning kernel reducing 128-element blocks across four
accumulators disagrees with a committing kernel that reduces once, by a few ULP
once d >= 128, and it abandoned rows that were genuinely in band: 1.24% of rows
placed within one ULP of the cut at d = 128, 1.65% at d = 256, 6.06% at d = 768.
Covering that took a widening term over a relative and an additive error source
with the second amplified by the first. One accumulation removes the divergence
instead, and removes the second pass over an admitted candidate with it: the old
shape traversed a candidate that was not abandoned twice, once to prune and once
to obtain the distance.

core/tests/range_search.rs::early_abandon_keeps_rows_just_inside_the_upper_cut
is the guard, at d = 768 with the cut one ULP above each probed row's own
distance. Restoring the two-kernel scan while keeping the raw cut makes it fail,
which is what distinguishes it from a formality.

Probe selection reuses kmeans::find_topk_batch, which recomputes every
selected centroid's distance with the direct kernel wherever its error bound
leaves the ranking ambiguous. Range search needs that: a query must select
the same lists alone or batched, or batch size becomes observable query
semantics. A focused kmeans test pins the property, and the IVF-Flat suite
guards it end to end.

Caller-bug checks are hoisted ahead of family dispatch, so an unsupported
family cannot mask a bad nprobe, a malformed filter or a metric mismatch as
a capability gap.

The probe width is a plain nprobe. An automatically widening mode is absent
rather than present-but-unsupported: a variant every well-formed value of which
only ever returns Unsupported is public surface with no working use. The params
struct keeps its fields private, so such a mode can arrive later as an added
constructor without breaking new().

Part of #97, step 1. The compressed families, cosine and inner product,
result caps and the bindings each follow separately.


jerry-024:
New: 3; still open: 0; fixed candidates: 0.

docs/api.html
```diff
@@ -74,6 +74,32 @@ <h2>Shared search parameters</h2>
...
+          <div class="table-wrap"><table><thead><tr><th>Field</th><th>Length</th><th>Description</th></tr></thead><tbody>
+            <tr><td><code>lims</code></td><td><code>query_count + 1</code></td><td>Offsets: query <code>i</code> spans <code>lims[i]..lims[i+1]</code>.</td></tr>
+            <tr><td><code>labels</code></td><td>Total rows</td><td>Row IDs, in no promised order.</td></tr>
+            <tr><td><code>distances</code></td><td>Total rows</td><td>Distances in the index's own space, so squared for <code>l2</code>.</td></tr>
+            <tr><td><code>query(i).stats</code></td><td>Per query</td><td><code>lists_probed</code>, <code>rows_scanned</code> (includes rejected and early-abandoned rows), <code>rows_committed</code>, <code>early_abandoned</code>, and <code>stop_reason</code>.</td></tr>
+            <tr><td><code>call_stats()</code></td><td>One per call</td><td><code>list_reads</code>, <code>list_rereads</code>, <code>passes</code>, and <code>peak_pending_result_rows</code>. These are call-level because a per-query peak cannot establish a call-wide concurrent peak.</td></tr>
```
jerry-024, line 98: **[MINOR] Documentation promises unavailable call statistics**

`RangeSearchCallStats` exposes only `list_reads`, but this table also lists `list_rereads`, `passes`, and `peak_pending_result_rows`. Code copied from the documentation therefore does not compile. Please document only `list_reads` for this release and add the other counters when they are implemented.

core/src/ivfflat_io.rs
```diff
@@ -1057,17 +1346,20 @@ fn scan_flat_rows(
...
-                }
+            let threshold = collector.cutoff();
+            // An infinite cutoff can never abandon a row, so entering the
+            // kernel would only cost a wasted SIMD pass over it.
+            if threshold.is_finite() && fvec_l2sqr_scaled_exceeds(query, vector, 1.0, threshold) {
+                collector.note_abandoned();
+                continue;
             }
             fvec_l2sqr(query, vector)
         } else {
```
jerry-024, line 1357: **[MAJOR] Avoid evaluating admitted candidates twice**

With a finite cutoff, every candidate that is not abandoned by `fvec_l2sqr_scaled_exceeds` has already traversed the full vector, then `fvec_l2sqr` traverses it again to obtain the distance. Broad or high-hit-rate range searches can therefore nearly double the distance work in this hot loop. Please return the completed distance from the cutoff kernel when it does not abandon, while preserving the existing accumulation and rounding contract.

core/src/range.rs
```diff
@@ -0,0 +1,1152 @@
...
+        max_width: usize,
+    },
+}
+
+/// Range search parameters. **Fields are private**: knobs introduced by later
+/// work would break struct-literal construction if they were `pub` fields, while
+/// declaring them now without implementing them would mean accepting a silently
+/// ineffective parameter. Private fields plus `new()` and `with_*` remove that
+/// dilemma.
+#[derive(Debug, Clone, Copy)]
```
jerry-024, line 590: **[MINOR] Do not expose an `Auto` mode that cannot succeed**

Every well-formed `RangeSearchWidth::Auto` value passes field validation only to return `Unsupported`, so the public variant and `with_width` path currently have no successful use. Since the enum is already `non_exhaustive`, please expose only the working fixed-`nprobe` constructor now and add `Auto` when automatic widening is implemented.

core/src/ivfflat_io.rs
JunRuiLee: Fixed, but not by returning the cutoff kernel's total -- it accumulates differently from fvec_l2sqr, so that value would change the distances we report.
New distance::fvec_l2sqr_unless_exceeds is fvec_l2sqr's own accumulation, probed every 128 elements: admitted rows are traversed once and the returned bits are unchanged. Two knock-ons: the abandon cut needs no widening any more, and top-K now prunes exactly, fixing a latent defect where it compared a differently
rounded sum against an fvec_l2sqr cutoff. No perf number -- my machine's noise is ~19%.

core/src/range.rs
JunRuiLee: Removed, plus `with_width` and `RangeSearchWidth` itself -- with only `Fixed` left the enum reaches no public signature. Fields stay private, so a widening mode can arrive later as an added constructor.

docs/api.html
JunRuiLee: Fixed. The table documents only `list_reads` now.

JunRuiLee:
@jerry-024 

On how the API is meant to grow -- the shape is deliberate, so it is worth
writing down what it was measured against.

**What I looked at.** FAISS declares `range_search` on `Index` and takes a
single `radius`. Doris and OpenSearch likewise expose one threshold. Knowhere
takes two (`radius` + `range_filter`). None of them takes SQL operators into
the index layer.

**A two-sided band, not a single radius.** A one-sided API cannot express a
band at all without two calls and a difference, while the second bound costs
nothing when it is `Unbounded` -- which is how the single-threshold case is
written here. So the two-sided form is strictly more expressive at no cost to
the common query, and Knowhere is the precedent that it is not exotic.

**Operators stay out of the index layer.** `DistanceBand` takes cuts, not
`<`/`<=`, so engine-specific vocabulary does not reach the kernels. Callers who
think in predicates get `DistanceEndpoint` + `CutOperator` as a separate
derivation step rather than as the search signature.

**The result container follows FAISS.** `lims` / `labels` / `distances` is the
shape of `RangeSearchResult`, because the row count per query is not known in
advance. Unsorted and unpadded, because ordering is the caller's business.

**Two error classes, because the difference is actionable.** `InvalidInput`
means the call is a bug worth fixing; `Unsupported` means this request cannot
be served and the caller should fall back to an exhaustive scan.

**The rule for growing it** is the one your `Auto` comment states, and I have
now applied it three times: nothing ships that no well-formed call can reach.
`RangeSearchWidth`, `limit_reached` and `StopReason`'s two unreachable variants
are all gone, each to return with the work that produces it. The params and
stats structs keep private fields so that stays additive.

**Two things the later steps have to answer**, and my current answer to both is
documentation rather than a signature change:

- Metric direction. `admit` is `lower <= value < upper` on the index's own
  value. Under L2 the upper bound is the natural cut; under inner product,
  where larger means nearer, it is the lower one. Both are expressible, but the
  half-openness is fixed on the upper side and does not follow the metric, so
  that needs saying explicitly when cosine and IP land.
- Strictness on the estimator families. `from_endpoints` turns `<= t` into
  `< nextafter(t)`, which is exact for IVF-Flat. For a quantized family the
  estimate's own error dwarfs that nudge, so the operator is honoured against
  the estimate, not against the true distance. The docs should say that plainly.

Happy to be talked out of any of it -- the bindings step is where these freeze,
so before then is the cheap time.

core/src/distance.rs
```diff
@@ -173,6 +253,25 @@ fn fvec_l2sqr_scalar(a: &[f32], b: &[f32]) -> f32 {
...
+    target_arch = "x86_64",
+    not(any(target_arch = "x86_64", target_arch = "aarch64"))
+))]
+#[inline]
+fn fvec_l2sqr_unless_exceeds_scalar(a: &[f32], b: &[f32], threshold: f32) -> Option<f32> {
+    let mut sum = 0.0f32;
+    for i in 0..a.len() {
+        let d = a[i] - b[i];
+        sum += d * d;
+        if sum > threshold {
```
jerry-024, line 273: [Minor] Match the scalar cutoff probe cadence to SIMD

This fallback checks `sum > threshold` after every coordinate, adding a compare and branch to the hot loop even though the SIMD kernels deliberately probe only every `L2_PROBE_STRIDE` elements. It is used on x86_64 hosts without AVX2 and on architectures without a dedicated SIMD implementation. Please probe at stride boundaries and once after the loop, as the AVX2/NEON paths do; L2 partial sums are monotonic, so this preserves correctness while avoiding the per-element branch.

JunRuiLee: Fixed in `27ab132`. The scalar fallback now accumulates in `L2_PROBE_STRIDE`-sized blocks, probes only at stride boundaries, and performs one final check after the remaining tail. This preserves the strict `>` and NaN behavior while removing the per-coordinate compare/branch. Verified with `cargo fmt --check`, native and x86_64 distance tests (17 each), and the range-search integration tests (31).

core/src/ivfflat_io.rs
```diff
@@ -763,6 +765,282 @@ impl<R: SeekRead> IVFFlatIndexReader<R> {
...
+                list_to_queries[list_id].push(qi);
+            }
+        }
+
+        // Per-query output buckets. The whole batch of (list, query) results is
+        // deliberately **not** materialized and merged afterwards: unlike a
+        // top-K local heap, a range collector has no upper bound, so that shape
+        // would hold O(B x Q x hits) resident at once. Each (list, query) task
+        // holds only that list's hits and takes the lock once to merge them.
+        let outputs: Vec<Mutex<Vec<(i64, f32)>>> =
```
jerry-024: Minor: Each `(list, query)` merge acquires `tallies[qi]` and `outputs[qi]` separately. Could we keep `(rows, scanned, early_abandoned)` behind one per-query mutex instead? This removes one mutex vector and one lock acquisition per merge without changing the parallelism or result semantics.

JunRuiLee: Fixed in `be9567f`. `(rows, scanned, early_abandoned)` now live in one `QueryMerge` behind a single per-query mutex. Both the streamed-list and regular scan paths use the same `merge_collector` closure, so each `(list, query)` merge takes one lock. Verified with `cargo fmt --check`, all 31 range-search integration tests, and the full core suite (541 passed, 1 ignored).

jerry-024 (approved):
+1

PR #104, opened 2026-09-15
feat: add IVF-RQ distance range search

JunRuiLee:
## Summary

Implements the IVF-RQ step of #97 on top of #99 (`75e391d`).

- Add L2 distance range search to both `VectorIndexReader` and `IVFRQIndexReader`, covering single/batch queries with and without a serialized Roaring allow-list.
- Reuse `Collector`, `RangeCollector`, `DistanceBand`, `VectorRangeSearchParams`, and `RangeSearchResult`.
- Decide membership using the one-bit estimate or the full multi-bit estimate, with half-open bounds, no clamping, no fixed-K truncation, and no ordering guarantee.
- Read each unique non-empty probed list once per call, share its payload across queries, and parallelize independent list/query scans. Return result-owned statistics without changing the last top-K statistics.
- Update the Rust/core and site documentation with estimator semantics, validation, filtering, and statistics.

## Correctness and compatibility

- Range scanning evaluates all stored bit planes using F32 lookup sums. It does not apply top-K's coarse lower-bound or quantized FastScan pruning to estimate-band membership.
- The range-only probe path checks every direct query-centroid distance before selection, including distances to unselected lists. Selected distances are reused in estimation, and ties are resolved by list ID.
- Non-finite centroids, consumed estimate factors, and computed distances return `InvalidData`. Invalid queries, widths, metrics, and filters retain the shared validation contract.
- Probing every list removes the IVF coverage gap, not quantization-induced missing or extra rows relative to an exact-distance predicate.
- Existing IVF-RQ top-K scheduling and scan code are unchanged. There are no storage-format changes.

This is a focused implementation on current main, not a wholesale transplant of the old `range-search-pr2`. Auto width, limits/caps, benchmark/CI tooling, other families/metrics, and C/JNI bindings are out of scope.

## Tests

Adds 13 tests covering an independently accumulated estimated-distance oracle (1–8 bits, padded dimensions and tail blocks), exact boundary membership, single/batch/filter equivalence, signed and large row IDs, fixed probe width, statistics and read capabilities, parallel scans, top-K compatibility, and non-finite data/arithmetic.

- `cargo fmt --all -- --check` — passed
- `git diff --check` — passed
- `cargo test -p paimon-vindex-core --lib --quiet` — 544 passed, 1 existing ignored
- `cargo test -p paimon-vindex-core --test range_search --quiet` — 41 passed
- Local links and anchors in the four changed HTML pages — passed


core/src/ivfrq_io.rs
```diff
@@ -609,6 +613,217 @@ impl<R: SeekRead> IVFRQIndexReader<R> {
...
+                };
+                if nprobe < distances.len() {
+                    distances.select_nth_unstable_by(nprobe - 1, compare);
+                    distances.truncate(nprobe);
+                    distances.shrink_to_fit();
+                }
+                distances.sort_unstable_by(compare);
+                Ok(distances)
+            })
+            .collect::<io::Result<Vec<_>>>()?;
```
jerry-024, line 725: **[MAJOR] Avoid materializing every centroid distance per query**

This batch path builds an nlist-sized (usize, f32) vector for every query before truncating it to nprobe, so it retains O(nq * nlist) tuples and performs scalar O(nq * nlist * d) work before any inverted-list I/O. At 100 queries and 65,536 lists, the tuples alone occupy about 100 MiB on a 64-bit target. Please reuse kmeans::find_topk_batch_with_centroid_norms, as the existing top-K path does, and pass its coarse distances into RQQueryTerms.

JunRuiLee: Thanks for calling this out. Addressed in 7a9547e18ccba82ee6b2ae5d9e4aeb01f75dbab5, now pushed to this PR.

The range path now calls `kmeans::find_topk_checked`, which keeps at most `min(nlist, 2 * nprobe)` centroid candidates. It periodically partitions that buffer back to `nprobe` using the existing distance-then-list-ID ordering, rather than collecting every centroid distance. Final sorting is in-place (`sort_unstable_by`), so wide probes do not allocate another sorting buffer. It still evaluates and validates every direct L2 distance before discarding a candidate, and passes the selected distances into `RQQueryTerms`. Query-level Rayon parallelism and the SIMD distance kernel are preserved; existing top-K behavior is unchanged. The buffer bound applies only to centroid selection, not to the number of returned range rows.

A local allocation probe with 65,536 lists and `nprobe=16` measured the selector's largest allocation dropping from **1,048,576 bytes to 512 bytes** (not a claim about total reader memory). Peak-live-allocation checks also pass for wide and full probes, including 1,024 lists with `nprobe=512` and `nprobe=1024`.

I did not directly substitute `find_topk_batch_with_centroid_norms` for two reasons:

1. **Error semantics:** that helper can discard a non-finite distance when selecting top-K. For example, two queries `[0]`, centroids `[0]` and `[1e20]`, and `nprobe=1` return the finite selected distance without an error. This PR explicitly requires `InvalidData` for overflow even in an unselected centroid. The checked selector preserves that contract without adding another full distance pass.
2. **Performance is workload-dependent:** in a local ARM64 probe-only benchmark (`nq=100`, `nlist=65536`, `nprobe=16`, 12 Rayon threads; nine-run medians with the three implementations interleaved in one process), the old/new paths took **15.7/14.3 ms at d=128** and **155.2/153.8 ms at d=768**. The existing batch helper took about **44.5/174.6 ms**, respectively. It *was* faster with one Rayon thread, so this is not a claim that direct scanning always wins; the bounded change fixes candidate memory without unconditionally switching the execution strategy.

Validation: the full core test suite passes, including **548 library tests and 42 range-search tests**, with existing ignores unchanged. Added coverage checks bounded candidate capacity, repeated compaction, ties, probe widths, large offsets, and multi-list single/batch/filter equivalence; the unselected-overflow regression now also covers 65 lists. Formatting and workspace/all-target Clippy with `-D warnings` pass.

jerry-024 (approved):
+1

PR #105, opened 2026-09-15
fix: upgrade crossbeam-epoch to address RUSTSEC-2026-0204

jerry-024:
Main currently locks `crossbeam-epoch` to 0.9.18, which is affected by [RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204.html). Upgrade it to the patched 0.9.20 release and synchronize the four Rust dependency license inventories.

Validation:
- `cargo +1.98.0 check --workspace --locked --offline`
- `cargo +1.98.0 fmt --all -- --check`
- `python3 tools/verify_binary_artifact.py --source`
- Verified the cached crate checksum and that each inventory changes only the dependency version.

Local test execution was blocked by unavailable execution approval; workspace tests remain to be verified by CI.


yuzelin (approved):
+1

PR #106, opened 2026-09-16
feat: add IVF-SQ L2 distance range search

JunRuiLee:
## Summary

Part of #97. Adds IVF-SQ **L2** distance range search on top of current `main`, including the already-merged IVF-RQ support from #104.

- Support single and batch queries, each with or without a serialized Roaring allow-list, through both the unified and direct SQ readers.
- Reuse `DistanceBand`, `VectorRangeSearchParams`, `RangeSearchResult`, and the generic `Collector` scan seam. Membership is decided by the same blocked SQ-estimated squared-L2 values used by the SQ scanner, not original-vector distances or a capped top-K fallback.
- Preserve the existing top-K behavior, other metrics, bindings, and on-disk format. Existing RQ range support remains enabled.
- Document quantization-induced membership errors, half-open bounds, statistics, and the supported-family matrix.

## Performance

- Retain blocked SIMD evaluation and finite-upper-cut early abandonment.
- Read each unique probed list once per batch, reuse the existing partition cache, and stream oversized lists in bounded chunks.
- Reuse scan scratch and query-owned collectors; large single queries merge once per list rather than once per row.
- Build compact, query-local filter masks once per list/chunk and share them across queries. Masks are not retained in the partition cache; fully excluded lists/chunks skip distance evaluation.

## Coverage

- SQ-estimate oracle across all four entry points and multiple probe widths.
- Exact boundary inclusion/exclusion, unbounded and empty bands, no result cap, and a concrete distinction from original-vector membership.
- SIMD blocks/tails, query permutation and parallel scans, cache reuse, query-local filtering, partial-probe deduplication, and bounded multi-range reads.
- Oversized-list single/batch/filter paths, invalid queries and metrics, non-finite estimates, payload errors, and fallible streaming.
- A lookup-count regression test proves that three queries over 1,024 rows perform 1,024 filter lookups instead of 3,072.

## Validation

Executed locally after rebasing onto `c2d9436`:

- `cargo test -p paimon-vindex-core --lib`: 553 passed, 1 existing ignored test.
- `cargo test -p paimon-vindex-core --test range_search`: 52 passed, including the existing Flat and RQ cases.
- `cargo test -p paimon-vindex-core --release --test range_search`: 52 passed.
- `cargo test -p paimon-vindex-core --test storage_format_fixtures`: 2 passed, 1 existing ignored fixture printer.

The exact CI Clippy invocation also supplies `-A clippy::chunks-exact-to-as-chunks`, which local Rust 1.95.0 does not recognize. The workspace Clippy command above passes without that lint allowance.

Linux/x86_64 and full cross-language CI remain for the repository workflows. There is no storage migration or binding change; reverting this additive PR does not require rewriting index files.

## Known review follow-up

Independent review found no blocking issues. One non-blocking P2 hardening follow-up remains: SQ range coarse-probe selection uses the existing unchecked centroid ranking, so a non-finite centroid or an overflowed coarse squared distance in an unselected list can escape validation. This is a source-traced corruption/extreme-value case, not a newly reproduced ordinary-input failure. A follow-up should add SQ-range-only checked probe validation and tests for unselected NaN/overflow cases; existing top-K and RQ behavior remain unchanged here.


core/src/ivfsq_io.rs
```diff
@@ -733,6 +746,228 @@ impl<R: SeekRead> IVFSQIndexReader<R> {
...
+                            ids,
+                            codes,
+                            &centroid,
+                            &sq,
+                            MetricType::L2,
+                            selection,
+                            &mut scratch,
+                            &mut collectors[query_index],
+                        )?;
+                    }
```
jerry-024, line 897: **[Major]** The oversized-list batch path serializes the per-query SQ scans.

For every streamed chunk, this loop walks all entries in `list_to_queries[first_list]` sequentially with a single `SqScanScratch`, while the normal-list path parallelizes the same per-query work. Because this branch only handles list payloads above 64 MiB, multiple queries probing the same list add substantial CPU time serially and leave Rayon workers idle. Please keep the shared chunk read, but parallelize the per-query scans with worker-local scratch and query-local collectors, then merge the chunk results.

core/tests/range_search.rs
```diff
@@ -947,6 +952,596 @@ fn serialize_roaring(allowed: &HashSet<i64>) -> Vec<u8> {
...
+                valid_filter.clone(),
+                if metric == MetricType::L2 {
+                    None
+                } else {
+                    Some(Unsupported)
+                },
+            ),
+        ] {
+            let params = VectorRangeSearchParams::new(band, nprobe);
+            for low_level in [false, true] {
```
jerry-024: **[Minor]** This test repeats the same validation rules across every wrapper.

The unified and direct entry points ultimately exercise the same batch range-search validation, so the metric × reader × entry-point matrix rechecks the dimension, non-finite, `nprobe`, and metric errors many times. Could we keep one invalid-input matrix against the direct batch path, cover malformed filters through the filtered path, and retain one delegation smoke case for the remaining wrappers? That would preserve the boundary coverage while removing most of this ~140-line test.

core/src/ivfsq_io.rs
JunRuiLee: Addressed in `eff3d50`.

Kept the shared chunk read and filter mask, and parallelized active queries with worker-local `SqScanScratch` and query-local collectors. Chunk results and counters are merged after the parallel scan completes; the single-query path and small-tail fallback remain sequential.

Added a regression that checks actual multi-worker execution, query mapping, rows and statistics, plus parallel collector-error propagation. The oversized-list integration now runs with both one and four threads. Formatting, workspace Clippy, 553 core unit tests, and all 52 range integration tests in debug and release pass locally.

core/tests/range_search.rs
JunRuiLee: Addressed in `eff3d50`.

Consolidated the invalid-input matrix into the direct batch path, checked malformed filters through the filtered batch path, and kept one invalid-dimension delegation smoke case per remaining wrapper. This removes the reader/entry-point cross product while retaining empty-band validation ordering and unsupported-metric coverage.

All 52 range integration tests pass in both debug and release.

jerry-024 (approved):
+1

PR #110, opened 2026-09-21
feat: add range search across language bindings

JunRuiLee:
## Summary

Expose the core range-search functionality from #108 through the existing C ABI/generated header, C++, JNI/Java, and Python bindings.

- Support IVF-Flat, IVF-SQ, IVF-PQ, and IVF-RQ with L2, cosine, and inner product, for single/batch queries with or without a serialized Roaring allow-list.
- Delegate raw half-open distance bands and public endpoint conversion to core, including structural unboundedness, strict/inclusive boundaries, and inner-product direction reversal.
- Return variable-length CSR results with per-query statistics, call-level list reads, and reader capability queries.
- Add explicit native-result ownership, exception-safe cleanup, checked lengths/conversions, and validation/error propagation.
- Preserve existing top-K APIs and behavior. Python still imports and executes top-K against older/partial native libraries; unavailable range APIs report a clear upgrade error.

## Tests and documentation

- Add a core-only oracle generator and 168 shared cases consumed independently by C, C++, Java, and Python. Cases cover the full family/metric/API matrix, bounded/unbounded/empty bands, and absent/non-empty/empty filters.
- Compare per-query label/distance-bit multisets, CSR offsets, and all counters without imposing an ordering guarantee absent from core.
- Add boundary, overflow, malformed-input, ownership, callback/reentry, old-library compatibility, and unaligned-NumPy-input regression tests.
- Document each language API and ownership contract; wire oracle generation and consumption into all four binding CI jobs.

## Validation

Validated locally on macOS:

- `cargo test --workspace`: **676 passed, 2 ignored**
- C11/C++17 strict builds (`-Wall -Wextra -Werror`), existing and new binding tests, and ASan/UBSan runs: passed; **168 oracle cases per binding**
- Java Maven API tests, native validation/panic-boundary/handle-safety tests, and the range oracle under `-Xcheck:jni`: passed; **168 oracle cases**
- Python tests with `PVI_RANGE_FIXTURES` enabled: **497 passed**, including **168 oracle cases**
- ASF license-header checks and `git diff --check`: passed

Independent pre-PR reviews were completed; the older-native-library and unaligned-input findings were fixed and re-reviewed. C++/JNI allocation-failure paths were reviewed but are not exhaustively fault-injected. Cross-platform CI remains to run on this PR.

## Scope and compatibility

Based on main at `0e25175` (#108). No core source/tests, storage-format, algorithm, or performance-tuning changes are included. DiskANN remains unsupported for range search. No data migration is required; rollback is a revert of this PR.


java/src/main/java/org/apache/paimon/index/vector/VectorRangeSearchResult.java
```diff
@@ -0,0 +1,145 @@
...
+            long[] labels,
+            float[] distances,
+            long[] lims,
+            long[] listsProbed,
+            long[] rowsScanned,
+            long[] rowsCommitted,
+            long[] earlyAbandoned,
+            long listReads) {
+        this.labels = Objects.requireNonNull(labels, "labels").clone();
+        this.distances = Objects.requireNonNull(distances, "distances").clone();
```
jerry-024: **[Major] Avoid copying every JNI-created result array twice**

JNI has already allocated and populated all seven Java arrays for each range-search result, but this constructor clones them again. Because range search has no top-K cap, wide bands or large batches pay another O(hit_count) copy and nearly double the Java-side peak result memory, which can turn an otherwise valid result into an OOM. Please keep the public constructor defensive, but add a package-private ownership-taking path for VectorIndexNative/JNI-created arrays.

include/paimon_vindex.hpp
```diff
@@ -669,6 +815,21 @@ class Reader {
     }
```
jerry-024, line 824: **[Minor] Remove the duplicate wrapper-side range validation**

All four C++ range-search entry points call this helper, which performs an extra metadata FFI call and repeats the overflow, length, and null checks already enforced by the shared C ABI path in ffi/src/range.rs. Keeping both paths adds a boundary call and creates two validation sources that can drift. Please delete this helper and its four call sites, keep the raw result initialized to null, and rely on the existing C ABI error plus check/copy_range_result path.

java/src/main/java/org/apache/paimon/index/vector/VectorRangeSearchResult.java
JunRuiLee: Thanks, fixed in d4a2d3c. JNI now uses a package-private ownership-taking factory; public construction and getters remain defensive. Added regression coverage.

include/paimon_vindex.hpp
JunRuiLee: Thanks, fixed in d4a2d3c. Removed the helper and all four calls, retaining null initialization and C ABI validation. Added filtered-input regression coverage.

java/src/main/java/org/apache/paimon/index/vector/VectorRangeSearchResult.java
```diff
@@ -0,0 +1,193 @@
...
+            throw new IllegalArgumentException("listReads must be non-negative");
+        }
+        this.listReads = listReads;
+    }
+
+    public int queryCount() {
+        return lims.length - 1;
+    }
+
+    public long[] labels() {
```
jerry-024, line 146: **[Major] Add a zero-copy way to consume range results**

This result is explicitly uncapped, but every public read path copies its payload: `labels()`, `distances()`, and `lims()` clone whole arrays, while the per-query accessors copy slices. A wide range can therefore fit when JNI creates the owned result yet fail with OOM as soon as the caller reads labels and distances, because the original arrays remain live and another full payload is allocated. This is separate from the fixed constructor-copy issue. Please retain the defensive methods if needed, but add indexed or read-only zero-copy accessors for labels, distances, and query offsets.

JunRuiLee: Fixed in 29b9d14. Added zero-copy indexed access to labels, distances, query bounds and counters, keeping defensive getters. Also bounded JNI staging and added a 2M-hit small-heap regression.

java/src/main/java/org/apache/paimon/index/vector/VectorRangeSearchResult.java
```diff
@@ -0,0 +1,242 @@
...
+        return labels.clone();
+    }
+
+    public long labelAt(int hitIndex) {
+        return labels[hitIndex];
+    }
+
+    public float[] distances() {
```
jerry-024: **[Major] Make raw distance semantics explicit across bindings**

`fromEndpoints` accepts public-space values, but range results expose core values as plain `distances` (`L2` is squared and inner product is negated). A caller that compares a returned value with the original endpoint therefore gets the wrong answer. Please apply one explicit contract across Rust, C, C++, Java, and Python: name result values `rawDistances`/`raw_distances` (including indexed accessors), and make raw-band constructors internal or explicitly raw-named so the public construction path is `fromEndpoints`.

JunRuiLee: Thanks, fixed in 3cdd045. Unified `raw_distances`/`rawDistances` (including indexed/query access) across all five languages, with explicit raw constructors and public endpoint factories. Added API-contract/endpoint regressions and updated docs; numeric values and zero-copy behavior are unchanged.

jerry-024 (approved):
+1

